PR #35, opened 2024-03-10
electron VFP in 1D

joglekara:
Just trying to reproduce Epperlein Haines for now.

It is kind of sort of there but not quite. 

The open Qs are - 
1. Why isn't the nuei fix for f1 working? We seem to need nuee to get any fidelity
2. Then, with nuee, we seem to need a scaling factor of 4*pi/3 / Z?
3. Any and all of these errors might just be from the kappa calculation. It includes a nuei which may or may not be correct.


joglekara:
But I am going to merge this because I would like to have the collision frequency calculations for the 1D2V Vlasov

PR #56, opened 2024-07-25
Update test.yaml to not separately lint utils

lazyoracle:
- `utils` was moved to `adept/`
- tests were not updated and are [failing](https://github.com/ergodicio/adept/actions/runs/10086093499)


joglekara:
closing because I caught this elsewhere already, but good catch! TY!

PR #91, opened 2024-12-05
Implemented an inialisation of local form of f10 by taking an artific…

jpbrodrick89:
Implemented an initialisation of local form of f10 by taking an artificially large timestep while keeping f0 fixed. This is essentially equivalent to temporarily reducing the electron mass and thus the effect of electron inertia. I have tested this in an untracked adept notebook and seems to work well when using nuei_scaling. It is untested when including anisotropic ee collisions, but should work as well as the assumption that a tridiagonal representation of these is sufficient. If this turns out to not be the case we might have to use some iterative method or a dense matrix solve. Currently, this is hard-coded and should ideally be an option in the yaml file but I did not want to create an extra dependency in the yaml file and break any of your code, if you can advise how to add switches while maintaining backwards compatibility going forwards that would be much appreciated.

Also implemented an option to boost isotropic electron-electron collisions to reduce nonlocal effects and restore to Maxwellian quicker. Note that at low Z electron inertia effects may still be prominent as the boost not impact this. Again no switch has been provided and the relevant lines have been commented out.

Also moved anisotropic ee diag contributions in the self.ee if statement as there is no point calculating these if using nuei_scaling.

Note vector_field is no longer in sync with oshun.py, whereas previously they were identical. We should consider removing oshun.py.

Finally, a few minor comments/questions, some we have previously discussed.

Happy to take on any feedback, thanks! :-) JPB

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-jax[cpu]
+jax[cuda12]
 diffrax
 matplotlib
 scipy
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
-jaxlib==0.4.26+cuda12.cudnn89
-jax==0.4.26
+jax[cpu]
 diffrax
 matplotlib
 scipy
--- a/adept/vfp1d/base.py
+++ b/adept/vfp1d/base.py
@@ -25,6 +25,7 @@ class BaseVFP1D(ADEPTModule):
         Te = u.Quantity(self.cfg["units"]["reference electron temperature"]).to("eV")
         Ti = u.Quantity(self.cfg["units"]["reference ion temperature"]).to("eV")
         Z = self.cfg["units"]["Z"]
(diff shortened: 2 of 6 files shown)


requirements.txt
```diff
@@ -1,5 +1,4 @@
-jaxlib==0.4.26+cuda12.cudnn89
-jax==0.4.26
+jax[cpu]
```
joglekara, line 1: can you swap this to `jax[cuda12]` and put this in the `requirements-cpu.txt` file?

joglekara (changes requested):
nice first PR! 

small change in the requirements requested. I also use `black` as my linter. VSCode has an integration with it which lints the code every time you save the file. We should either add that as a pre-commit hook or have your VSCode integrate it or both. It'll just keep the styles the same

joglekara (approved):
Looks good to go

PR #96, opened 2025-04-08
New diagnostics / analysis for Vlasov1D

joglekara:
Added diagnostic outputs that calculate the change in the distribution function from the vlasov terms and from the FP terms separately. This is for use in downstream analysis.

These are stored by setting `cfg["terms"]["diags"] = True`

### DIFF
--- a/adept/_vlasov1d/helpers.py
+++ b/adept/_vlasov1d/helpers.py
@@ -243,16 +243,16 @@ def post_process(result: Solution, cfg: Dict, td: str, args: Dict):
                 plt.close()
 
     f_xr = store_f(cfg, result.ts, td, result.ys)
-    if cfg["terms"]["diags"]:
-        diags_xr = xarray.Dataset(
-            {
-                spc: xarray.DataArray(
-                    result.ys["diags"][spc],
-                    coords=(("t", result.ts["diags"]), ("x", cfg["grid"]["x"]), ("v", cfg["grid"]["v"])),
-                )
-                for spc in ["diag-vlasov", "diag-fp"]
-            }
-        )
+
+    diags_dict = {}
+    for k in ["diag-vlasov-dfdt", "diag-fp-dfdt"]:
+        if cfg["diagnostics"][k]:
+            diags_dict[k] = xarray.DataArray(
+                result.ys[k], coords=(("t", result.ts[k]), ("x", cfg["grid"]["x"]), ("v", cfg["grid"]["v"]))
+            )
+
+    if len(diags_dict.keys()):
+        diags_xr = xarray.Dataset(diags_dict)
         diags_xr.to_netcdf(os.path.join(td, "binary", "diags.nc"))
 
     mlflow.log_metrics({"postprocess_time_min": round((time() - t0) / 60, 3)})
--- a/adept/_vlasov1d/helpers.py
+++ b/adept/_vlasov1d/helpers.py
@@ -243,6 +243,17 @@ def post_process(result: Solution, cfg: Dict, td: str, args: Dict):
                 plt.close()
 
     f_xr = store_f(cfg, result.ts, td, result.ys)
+    if cfg["terms"]["diags"]:
+        diags_xr = xarray.Dataset(
+            {
+                spc: xarray.DataArray(
+                    result.ys["diags"][spc],
+                    coords=(("t", result.ts["diags"]), ("x", cfg["grid"]["x"]), ("v", cfg["grid"]["v"])),
+                )
+                for spc in ["diag-vlasov", "diag-fp"]
+            }
+        )
+        diags_xr.to_netcdf(os.path.join(td, "binary", "diags.nc"))
 
     mlflow.log_metrics({"postprocess_time_min": round((time() - t0) / 60, 3)})
 
--- a/adept/_vlasov1d/modules.py
+++ b/adept/_vlasov1d/modules.py
@@ -198,8 +198,6 @@ class BaseVlasov1D(ADEPTModule):
     def init_diffeqsolve(self):
         self.cfg = get_save_quantities(self.cfg)
         self.time_quantities = {"t0": 0.0, "t1": self.cfg["grid"]["tmax"], "max_steps": self.cfg["grid"]["max_steps"]}
-        # print(self.cfg["save"])
-
         self.diffeqsolve_quants = dict(
             terms=ODETerm(VlasovMaxwell(self.cfg)),
             solver=Stepper(),
--- a/adept/_vlasov1d/modules.py
+++ b/adept/_vlasov1d/modules.py
@@ -188,9 +188,9 @@ class BaseVlasov1D(ADEPTModule):
         for field in ["a", "da", "prev_a"]:
             state[field] = jnp.zeros(self.cfg["grid"]["nx"] + 2)  # need boundary cells
 
-        if self.cfg["terms"]["diags"]:
-            state["diag-vlasov"] = jnp.zeros_like(f)
-            state["diag-fp"] = jnp.zeros_like(f)
+        for k in ["diag-vlasov-dfdt", "diag-fp-dfdt"]:
+            if self.cfg["diagnostics"][k]:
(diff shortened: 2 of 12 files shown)


adept/_vlasov1d/modules.py
```diff
@@ -188,19 +188,16 @@ def init_state_and_args(self) -> Dict:
         for field in ["a", "da", "prev_a"]:
             state[field] = jnp.zeros(self.cfg["grid"]["nx"] + 2)  # need boundary cells
 
+        if self.cfg["terms"]["diags"]:
+            state["diag-vlasov"] = jnp.zeros_like(f)
```
johnbcoughlin, line 192: I think a more descriptive name might be `diag-vlasov-dfdt` and `diag-fp-dfdt`

adept/_vlasov1d/storage.py
```diff
@@ -74,43 +75,38 @@ def store_f(cfg: Dict, this_t: Dict, td: str, ys: Dict) -> xr.Dataset:
...
+    elif {"t", "kx", "v"} == set(cfg["save"]["diags"].keys()):
+        axes = (("t", this_t["diag-vlasov"]), ("kx", cfg["grid"]["kx"]), ("v", cfg["grid"]["v"]))
+    elif {"t", "x", "kv"} == set(cfg["save"]["diags"].keys()):
+        axes = (("t", this_t["diag-vlasov"]), ("x", cfg["grid"]["x"]), ("kv", cfg["grid"]["kv"]))
+    elif {"t", "kx", "kv"} == set(cfg["save"]["diags"].keys()):
+        axes = (("t", this_t["diag-vlasov"]), ("kx", cfg["grid"]["kx"]), ("kv", cfg["grid"]["kv"]))
+    else:
+        raise NotImplementedError
+
+    diags_store = xr.Dataset({spc: xr.DataArray(ys[spc], coords=axes) for spc in ["diag-vlasov", "diag-fp"]})
```
johnbcoughlin, line 102: You should make this pair of strings, `["diag-vlasov", "diag-fp"]`, an element of the config. That way you could reuse it across the places it appears, and do stuff like `if "diag-vlasov-dfdt" in cfg["diagnostics"]

adept/_vlasov1d/storage.py
```diff
@@ -136,21 +132,107 @@ def fields_save_func(t, y, args):
...
 
-    elif {"t", "x", "ky"} == set(cfg["save"][k].keys()):
+    return dist_save_func
+
+
+def get_diag_save_func(axes, dist_save_config, dist_key):
+    if {"t"} == set(dist_save_config.keys()):
+
+        def dist_save_func(t, y, args):
+            return {"diag-vlasov": y["diag-vlasov"], "diag-fp": y["diag-fp"]}
```
johnbcoughlin, line 182: same here, could do `{diag_name: y[diag_name] for diag_name in cfg["diagnostics"]}`

adept/_vlasov1d/storage.py
```diff
@@ -136,21 +132,107 @@ def fields_save_func(t, y, args):
...
+
+    elif {"t", "x", "v"} == set(dist_save_config.keys()):
+
+        def dist_save_func(t, y, args):
+            diags = {}
+            diags["diag-vlasov"] = interp2d(
```
johnbcoughlin, line 188: these two statements can be a for loop

adept/_vlasov1d/storage.py
```diff
@@ -136,21 +132,107 @@ def fields_save_func(t, y, args):
...
+                kind="linear",
+            )
+            return diags
+
+    elif {"t", "kx", "v"} == set(dist_save_config.keys()):
+
+        def dist_save_func(t, y, args):
+            diags = {}
+            fkx = jnp.fft.rfft(y["diag-vlasov"], axes=0)
+            diags["diag-vlasov"] = interp2d(
```
johnbcoughlin, line 212: same here

adept/_vlasov1d/storage.py
```diff
@@ -165,29 +247,39 @@ def get_save_quantities(cfg: Dict) -> Dict:
...
+        elif save_type.casefold() == "electron":
+            save_config["func"] = get_dist_save_func(
+                axes={"x": cfg["grid"]["x"], "v": cfg["grid"]["v"], "kx": cfg["grid"]["kx"]},
+                dist_save_config=save_config,
+                dist_key=save_type,
+            )
+
+        elif save_type.casefold() == "diags":
+            save_config["func"] = get_diag_save_func(
+                axes={"x": cfg["grid"]["x"], "v": cfg["grid"]["v"], "kx": cfg["grid"]["kx"]},
```
johnbcoughlin, line 275: use `axes = {dim: cfg["grid"][dim] for dim in ["x", "v", "kx"]}`

configs/vlasov-1d/bump-on-tail.yaml
```diff
@@ -67,6 +67,7 @@ terms:
   field: poisson
   edfdv: exponential
   time: sixth
+  diags: false
```
johnbcoughlin, line 70: I think it's better to use a list here which may be empty, see my other comments for how that can be useful.

adept/_vlasov1d/storage.py
```diff
@@ -74,43 +75,38 @@ def store_f(cfg: Dict, this_t: Dict, td: str, ys: Dict) -> xr.Dataset:
...
-#         store_f(cfg, this_t[-1:], td, running_f)
-#         mlflow.log_artifacts(td)
-#         clean_td(td)
-
-
-def get_field_save_func(cfg, k):
-    if {"t"} == set(cfg["save"][k].keys()):
+def store_diags(cfg: Dict, this_t: Dict, td: str, ys: Dict) -> xr.Dataset:
+    """
+    Stores f to netcdf
```
johnbcoughlin, line 80: this is wrong, it stores the diagnostics to netcdf

adept/_vlasov1d/modules.py
joglekara: agreed...

adept/_vlasov1d/storage.py
joglekara: I was hesitant to give each their own save config but it does allow for much cleaner practices inside

adept/_vlasov1d/storage.py
joglekara: combined this with `get_dist_save_func()`

adept/_vlasov1d/solvers/vector_field.py
```diff
@@ -155,14 +155,23 @@ def __init__(self, cfg: Dict):
         else:
             raise NotImplementedError
         self.fp = fokker_planck.Collisions(cfg=cfg)
+        self.vlasov_dfdt = cfg["diagnostics"]["diag-vlasov-dfdt"]
+        self.fp_dfdt = cfg["diagnostics"]["diag-fp-dfdt"]
```
johnbcoughlin, line 159: I support adding module-level fields like this so we don't have to reference `cfg` in the `call` methods.

adept/_vlasov1d/helpers.py
```diff
@@ -244,6 +244,17 @@ def post_process(result: Solution, cfg: Dict, td: str, args: Dict):
 
     f_xr = store_f(cfg, result.ts, td, result.ys)
 
+    diags_dict = {}
+    for k in ["diag-vlasov-dfdt", "diag-fp-dfdt"]:
```
johnbcoughlin, line 248: I still think it's worth making this list of two strings a well-known constant in some way, either as a yaml list in `cfg["diagnostics"]` or just a python module-level constant like `DFDT_DIAGNOSTICS`.

johnbcoughlin (approved):
LGTM, just one outstanding comment but not a blocker.

adept/_vlasov1d/helpers.py
joglekara: this probably applies to all configuration parameters in some way right? maybe we should leave it for a future, broader refactor

PR #97, opened 2025-04-12
Add a flag for logging artifacts in`call` and `val_and_grad` for ergoExo

joglekara:
Add a flag for logging artifacts in`call` and `val_and_grad` for ergoExo
Also removes mlflow export from run.py

### DIFF
--- a/adept/_base_.py
+++ b/adept/_base_.py
@@ -319,7 +319,7 @@ class ergoExo:
 
         return modules
 
-    def __call__(self, modules: Dict = None) -> Tuple[Solution, Dict, str]:
+    def __call__(self, modules: Dict = None, export=True) -> Tuple[Solution, Dict, str]:
         """
         This function is the main entry point for running a simulation. It takes a configuration dictionary and returns a
         ``diffrax.Solution`` object and a dictionary of datasets. It calls the ``self.adept_module``'s ``__call__`` function.
@@ -349,7 +349,8 @@ class ergoExo:
             t0 = time.time()
             with tempfile.TemporaryDirectory(dir=self.base_tempdir) as td:
                 post_processing_output = self.adept_module.post_process(run_output, td)
-                robust_log_artifacts(td)  # logs the temporary directory to mlflow
+                if export:
+                    robust_log_artifacts(td)  # logs the temporary directory to mlflow
 
                 if "metrics" in post_processing_output:
                     mlflow.log_metrics(post_processing_output["metrics"])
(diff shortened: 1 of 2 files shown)


jpbrodrick89:
Maybe we can have a chat about this. With `export=False` there is still a lot of logging happening in `_setup_` and post-processing. Would be good to understand why you would want one and not the other.

joglekara:
Yeah the intention here is to just save on logging artifacts because they can occupy a lot of storage (which costs more than just storing parameters and metrics in a DB). We could change the flag to `export_artifacts` instead to be more precise, and can implement the rest later, if needed.

adept/_base_.py
```diff
@@ -386,7 +387,8 @@ def val_and_grad(self, modules: Dict = None) -> Tuple[float, Dict, Tuple[Solutio
             t0 = time.time()
             with tempfile.TemporaryDirectory(dir=self.base_tempdir) as td:
                 post_processing_output = self.adept_module.post_process(run_output, td)
-                robust_log_artifacts(td)  # logs the temporary directory to mlflow
+                if export:
+                    robust_log_artifacts(td)  # logs the temporary directory to mlflow
                 if "metrics" in post_processing_output:
                     mlflow.log_metrics(post_processing_output["metrics"])
```
jpbrodrick89, line 393:
```suggestion
            post_processing_output = self.adept_module.post_process(run_output)
            if export_all:
                with tempfile.TemporaryDirectory(dir=self.base_tempdir) as td:
                    robust_log_artifacts(td)  # logs the temporary directory to mlflow
            if export_metrics and "metrics" in post_processing_output:
                    mlflow.log_metrics(post_processing_output["metrics"])
```
Personally, I'd find it more helpful if there was more customisation on what was exported. Metrics may be data-intensive and undesirable for logging. Sometimes, Mlflow might want to be removed entirely including on setup. We could have multiple flags `export_all, export_metrics, export_init, disable_mlflow`.

I think `adept_module.post_process` should also be refactored so that it no longer takes `td` as an input, forcing the logging to happen on a separate line. This may be a breaking change but I know that `lagradept` for example does not actually use the `td` argument.

Nevertheless, happy for you to merge this to use for now and we work in something more all encompassing afterwards.

run.py
```diff
@@ -31,6 +31,3 @@
     else:
         exo.run_job(args.run_id, nested=None)
         run_id = args.run_id
-
-    if "MLFLOW_EXPORT" in os.environ:
-        export_run(run_id)
```
jpbrodrick89, line 36: Where does this happen now?

jpbrodrick89 (approved):
Forgot to approve in previous review, comments are optional.

adept/_base_.py
jpbrodrick89: Saw you said something very similar in your last comment, that approach sounds great!

run.py
joglekara: it doesn't need to happen anymore because I figured out how to directly communicate to the parent MLFlow server

adept/_base_.py
joglekara: also should just be called something else, it really is only a logging wrapper

PR #98, opened 2025-05-01
chore: added pyproject.toml

jpbrodrick89:
I'm only just trying to understand these now, so not sure how to check this is correct. Feel free to edit/suggest changes.

My main aim here is to iteratively build repo foundations (both adept and lagradept) to make dev work smoother in particular linting with mandatory pre-commit hooks. I am doing this step by step to make review process simpler.


pyproject.toml
```diff
@@ -0,0 +1,86 @@
+[project]
+name = "adept"
+description = "Automatic-Differentiation-Enabled Plasma Transport in JAX"
+readme = "README.md"
+authors = [
+    {name = "Archis Joglekar/Ergodic LLC", email="archis@ergodic.io"},
+]
+license = "MIT"
+license-files = ["LICENSE*"]
+requires-python = ">=3.10"
```
joglekara, line 10: I just bumped to 3.11. can you do that here too?

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+        "numpy",
+        "tqdm",
+        "xarray",
+        "mlflow",
+        "flatdict",
+        "h5netcdf",
+        "optax",
+        "boto3",
+        "pint",
+        "mlflow_export_import",
```
joglekara, line 24: this is no longer needed and removed from the requirements in the most recent PR

joglekara (changes requested):
ty for doing this. some small changes

jpbrodrick89:
The only thing I'm not sure about is how to enforce that `cpu/gpu` is selected, it might be better to have `jax` in core dependencies and have the `cpu` requirements empty.

Also should we add `validate-pyproject` into CI?

joglekara:
- 

> The only thing I'm not sure about is how to enforce that `cpu/gpu` is selected, it might be better to have `jax` in core dependencies and have the `cpu` requirements empty.

I think the correct version is chosen if you do `pip install adept[cpu]` vs `pip install .[gpu]`. Does that answer your question? Either way, feel free to do what you think is best. There is no purpose for this file right now besides whatever it is you need it for!

> 
> Also should we add `validate-pyproject` into CI?
I have no idea what this is! Your call

jpbrodrick89:
What I mean is should it throw an error when we run install `pip install adept` or should this just default to be the equivalent to `pip install adept[cpu]`. I think I'll go with the latter which is what Jax does.

joglekara:
yes that sounds reasonable

.github/workflows/cpu-tests.yaml
```diff
@@ -4,33 +4,48 @@ name: cpu-tests
 
 # Controls when the action will run.
 on:
-  # Triggers the workflow on push or pull request events but only for the master branch
+  pull_request:
```
joglekara, line 7: do you know if forked repos will run the checks on a PR too? If so, this is probably unsafe, right?

.github/workflows/cpu-tests.yaml
```diff
@@ -4,33 +4,48 @@ name: cpu-tests
...
+
   push:
     branches:
       - main
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+# Cancel in-progress tests on new commits to the same branch
+concurrency:
```
joglekara, line 17: this is nice, ty

.github/workflows/cpu-tests.yaml
jpbrodrick89: Note also that just by editing the workflow in the pr and not committing I was able to trigger the tests. The same might be possible if pr'ing from a fork too.

joglekara: The worry is that anybody can make a fork and a PR back to `origin/main` while only authorized users can make a branch and make a PR. Would we be able to prevent the former?

joglekara: ah I just clicked on the link :) . Looks like we should put in an org name: repo name and then we're okay

jpbrodrick89: Wrt the first answer: I'm not 100% sure, but there's a chance the suggestion only users forks running the workflows on their own fork, but as I said not sure.

The second answer looks a lot more like what we want, requiring manual approval for outside collaborators. This might be of interest too: https://github.com/marketplace/actions/manual-workflow-approval

Finally, are our runners GitHub-hosted or self-hosted. If GitHub-hosted we should only really worry about cost rather than maliciious attacks but if self-hosted we need to be VERY careful (indeed it is recommended one should never use self-hosted for public repos): https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#hardening-for-self-hosted-runners

joglekara: It's GitHub hosted but I like to sometimes report the test results to MLFlow in case we need to diagnose them further. I don't think this is implemented yet in `adept` but is for `tsadar`.  If these run on their own fork's actions (which I am starting to suspect they do), that fork wouldn't have access to the MLFlow credentials anyway, so we might be safe.

In any case, I am open to changes as long as we can ensure that they will not be able to run actions within the main repo's environment. Seems like that should be the case? maybe we need to fork around and find out.

joglekara: I believe this is now resolved by changing the settings to always require an approval

joglekara (dismissed):
good to go now?

jpbrodrick89:
All good, feel free to hit squash and merge 🙂

jpbrodrick89:
Actually, need to update the file to match your new requirements will see if I have time tonight but feel free to do it on your side if you like.

jpbrodrick89:
Done, if we set up uv in the future, syncing all our various requirements files with pre-commit hooks becomes a lot simpler/actually possible!

PR #99, opened 2025-05-01
chore: add mlruns directory to gitignore

jpbrodrick89:
If simulations are run in the repo directory for testing these folders show up in `git status`. If we'd rather dissuade these being created in the repo in the first place, I'm happy to close.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,5 +125,8 @@ venv.bak/
 .dmypy.json
 dmypy.json
 
+# mflow
+*mlruns/
+
 # Pyre type checker
 .pyre/


joglekara:
we should probably just remove the `.gitignore` and let users manage their own gitignores. wdyt?

jpbrodrick89:
I tend to agree with the approach suggested [here](https://stackoverflow.com/questions/5765645/should-you-commit-gitignore-into-the-git-repos). We should definitely keep a .gitignore but if you feel mlruns is too specific I'll just add it in my own global config if you prefer.

joglekara:
interesting, in that case, `mlruns` is a very obvious and desirable add!

joglekara (approved):
lgtm!

PR #100, opened 2025-05-06
Fixing tests after diagnostics add

joglekara:


tests/test_vfp1d/test_kappa_eh.py
```diff
@@ -34,7 +36,8 @@ def _run_(Z, ee):
 @pytest.mark.parametrize("Z", list(range(1, 21, 4)) + [40, 60, 80])
 @pytest.mark.parametrize("ee", [True, False])
 def test_kappa_eh(Z, ee):
-    if "CPU_ONLY" in os.environ:
+    if not any(["gpu" == device.platform for device in devices()]):
+        pytest.skip(f"Skipping Z={Z} to save time because no GPU is available")
```
jpbrodrick89, line 40: This is the commit that removed the CPU tests by accident, so we know that any commits passing tests before this point and after my init-local-f10 commit would include this. Unfortunately, there are none of these and the logs are gone so I will have to do manually 🤦🏻

PR #101, opened 2025-05-09
Testing whether forks trigger CI without manual approval

jpbrodrick89:


jpbrodrick89:
Looks like the tests still run (and this is on main before my orig PR is merged). Can you check whether this runs on your actions rather than mine? Seems like right now anyone can edit workflows and have them run without manual approval. However, there is a chance that if I was a brand new user (not one that's contributed before), then this wouldn't run automatically.

I think the only way to address this is manual workflow approval and to lock the workflows file if possible.

jpbrodrick89:
Even more worryingly I can actually merge this PR without approval...

joglekara:
> Looks like the tests still run (and this is on main before my orig PR is merged). Can you check whether this runs on your actions rather than mine? 

Yes it runs in the repo's actions.

Seems like right now anyone can edit workflows and have them run without manual approval. However, there is a chance that if I was a brand new user (not one that's contributed before), then this wouldn't run automatically.
> 
> I think the only way to address this is manual workflow approval and to lock the workflows file if possible.

I think some (or all) of this is because you have write access to the repo. I would like to think others cannot do all this.

PR #104, opened 2025-05-24
refactor(test): set jax.config in conftest

jpbrodrick89:
All test files call `conftest.py` implicitly first meaning we only need to set the `jax.config` once. This was ported from the `ruff` PR for focus on this change. Apologies for polluting with other linting changes.


tests/test_tf1d/test_landau_damping.py
```diff
@@ -1,17 +1,16 @@
...
 #  research@ergodic.io
-import yaml
-
 import numpy as np
+import yaml
 from jax import config
 
 from adept import ergoExo
 
 config.update("jax_enable_x64", True)
```
joglekara, line 9: this can go right?

jpbrodrick89: I'll do a grep pass through and remove the lot, sorry for missing!

joglekara (approved):
🚢

PR #122, opened 2025-08-22
chore: add mlruns to gitignore

johnbcoughlin:


jpbrodrick89:
I think we already have `*mlruns/` in there, does this not catch it? Feel free to refactor to `**mlruns/` if that is universally supported to catch it if created in deep subdirectories.

jpbrodrick89:
@johnbcoughlin does this add anything or can we close?

PR #153, opened 2025-12-04
moving arrays to CPU before postprocessing to avoid GPU memory problems

joglekara:


johnbcoughlin:
LGTM, but seems like you might need to rebase to fix conflicts

PR #154, opened 2025-12-04
Vlasov1D SRS fixes

joglekara:
1. re-adds noise capability, we likely will need to add noise at each timestep like we do in LPSE, instead, though.
2. fixes a bug in the dt calculation due to units package


adept/_vlasov1d/modules.py
```diff
@@ -109,6 +109,7 @@ def get_solver_quantities(self) -> dict:
         :return:
         """
         cfg_grid = self.cfg["grid"]
+        print(cfg_grid)
```
johnbcoughlin: delete :)

PR #156, opened 2025-12-08
Chang-Cooper operator for Vlasov 1D

joglekara:
vibe coded a chang-cooper operator that conserves density to 1e-16 but energy conservation is worse than the regular implementation of LB and DG

Also added a test for density and energy conservation

### DIFF
--- a/adept/_vlasov1d/solvers/pushers/fokker_planck.py
+++ b/adept/_vlasov1d/solvers/pushers/fokker_planck.py
@@ -132,7 +132,7 @@ class _DriftDiffusionBase:
 
 
 class LenardBernstein(_DriftDiffusionBase):
-    """Classic Lenard-Bernstein Fokker-Planck operator. """
+    """Classic Lenard-Bernstein Fokker-Planck operator."""
 
     def __init__(self, cfg: Mapping[str, Any]):
         """


adept/_vlasov1d/solvers/pushers/fokker_planck.py
```diff
@@ -75,12 +80,109 @@ def __call__(
...
+        c = c.at[:, :-1].set(lam * beta)
+
+        beta_l = jnp.concatenate([jnp.zeros((self.cfg["grid"]["nx"], 1)), beta], axis=1)
+        alpha_r = jnp.concatenate([alpha, jnp.zeros((self.cfg["grid"]["nx"], 1))], axis=1)
+        diag = self.ones + lam * (alpha_r - beta_l)
+
+        return a, diag, c
+
+
+class ChangCooperDougherty:
```
johnbcoughlin: I am not familiar with the convention that

- Dougherty = `dv (T dv f + (v - u) f)`
whereas
- Lenard-Bernstein = `dv (T dv f + v f)` (i.e. u = 0).

I have always seen the nonzero mean velocity case described as Dougherty or L-B, interchangeably. I think you should make the zero mean velocity linearization explicit in the name of the class, or at the very least a comment.

johnbcoughlin: Speaking of which, there is an obvious code-sharing opportunity here -- why not have the `u=0` case pass a zero vbar/u array directly to this class?

tests/test_vlasov1d/test_fokker_planck_conservation.py
```diff
@@ -0,0 +1,171 @@
...
+    # Final distribution should be close to Maxwellian (within 10% L2 error)
+    # This is a loose bound since we're testing conservation primarily
+    assert relative_error < 0.1, (
+        f"{operator_type}: Distribution did not thermalize properly. L2 error to Maxwellian: {relative_error:.2e}"
+    )
+
+    print("  Thermalization test passed!")
+
+
+if __name__ == "__main__":
```
johnbcoughlin: remove this?

tests/test_vlasov1d/test_fokker_planck_conservation.py
```diff
@@ -0,0 +1,171 @@
...
+    Test that Fokker-Planck operators conserve density and energy.
+
+    Args:
+        operator_type: Either "Lenard_Bernstein" or "Dougherty"
+    """
+    with open("tests/test_vlasov1d/configs/fokker_planck_conservation.yaml") as file:
+        config = yaml.safe_load(file)
+
+    # Set the operator type
+    config["terms"]["fokker_planck"]["type"] = operator_type
```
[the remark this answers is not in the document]
joglekara: Ya its spatially uniform. both terms should be 0 because dx and E are 0.

adept/_vlasov1d/solvers/pushers/fokker_planck.py
johnbcoughlin: I'm still not convinced that DG/LB is widely understood as being the drift momentum split. Gkeyll calls it Lenard Bernstein everywhere, and their operator certainly includes drift velocity.

PR #157, opened 2025-12-10
Implement MlflowLoggingModule

johnbcoughlin:
Adds a new equinox module wrapper concept that optionally starts an mlflow run and passes the ID to pre_logging and post_logging methods.

### DIFF
--- a/adept/mlflow_logging.py
+++ b/adept/mlflow_logging.py
@@ -1,4 +1,4 @@
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from dataclasses import dataclass
 
 import equinox as eqx
@@ -28,7 +28,7 @@ class MLRunId(eqx.Module):
         return MLRunId("0" * 32)
 
 
-class AdroitModule(eqx.Module, ABC):
+class MlflowLoggingModule(eqx.Module):
     with_mlflow: bool
 
     @staticmethod
@@ -37,13 +37,12 @@ class AdroitModule(eqx.Module, ABC):
             with mlflow.start_run() as run:
                 return MLRunId(run.info.run_id)
 
-        return jax.experimental.io_callback(
-            create_mlflow_run, MLRunId.example(), dummy_arg, sharding=None, ordered=False
-        )
+        return jax.experimental.io_callback(create_mlflow_run, MLRunId.example(), dummy_arg)
 
-    def call_in_callback(self, callable, *args, mlflow_run_id, **kwargs):
+    @staticmethod
+    def call_logfunc_in_callback(logfunc, *args, mlflow_run_id, **kwargs):
         def call_with_byte_array_id(*args, mlflow_run_id, **kwargs):
-            callable(*args, mlflow_run_id=str(mlflow_run_id), **kwargs)
+            logfunc(*args, mlflow_run_id=str(mlflow_run_id), **kwargs)
 
         jax.debug.callback(call_with_byte_array_id, *args, mlflow_run_id=mlflow_run_id, **kwargs)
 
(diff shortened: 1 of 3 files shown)


johnbcoughlin:
Couple design points:

- `io_callback` can't depend on any of the args that you differentiate with respect to. That has a couple implications:
    - The start_run call has to be in a separate callback from pre_logging (assuming you want to pre-log some of the args)
    - We have to pass an explicit `jnp.arange` as `mlflow_batch_num` to ensure separate mlflow runs. It seems that vmap doesn't vmap over parameters that aren't specified at the call site, so I don't think there's a clever way to use a default value of the right size, or build one up inside of the function.

There are probably other options available, like `@custom_vmap`, but then you need to use `@custom_vjp` -> you can't use forward diff anyway. This design has the minimal amount of "magic" but should still allow the call patterns we want. If passing `mlflow_batch_num` turns out to be very painful in practice, we can revisit.

adept/adroit.py
```diff
@@ -0,0 +1,102 @@
...
+    with_mlflow: bool
+
+    @staticmethod
+    def create_mlflow_run_in_callback(dummy_arg):
+        def create_mlflow_run(dummy_arg):
+            with mlflow.start_run() as run:
+                return MLRunId(run.info.run_id)
+
+        return jax.experimental.io_callback(
+            create_mlflow_run, MLRunId.example(), dummy_arg, sharding=None, ordered=False
```
jpbrodrick89:
```suggestion
            create_mlflow_run, MLRunId.example(), dummy_arg,
```

Don't see why we want to explicitly provide the default kwargs? Or is this for documentation?

adept/adroit.py
```diff
@@ -0,0 +1,102 @@
...
+            self.pre_logging(*args, mlflow_run_id=str(mlflow_run_id), **kwargs)
+
+        jax.debug.callback(pre_logging_call, *args, mlflow_run_id=mlflow_run_id, **kwargs)
+
+    def call_post_logging_in_callback(self, result, *args, mlflow_run_id, **kwargs):
+        def post_logging_call(result, *args, mlflow_run_id, **kwargs):
+            self.post_logging(result, *args, mlflow_run_id=str(mlflow_run_id), **kwargs)
+
+        jax.debug.callback(post_logging_call, result, *args, mlflow_run_id=mlflow_run_id, **kwargs)
+
```
jpbrodrick89: Is the point of this just to ensure `pre/post_logging` functions have an intuitive API with a string run id? If so can we rename these to something more informative, e.g. `pre_log_from_byte_array_id`.

I'm guessing we can't do `str(mlflow_run_id)` in the main body as Jax would forbid that.

jpbrodrick89:
Yes, the name might be a bit cute 😅 ... or more to the point undescriptive 🙂 maybe just go with something really descriptive like `JittableLogger` or `CallbackLogger`. I really hate that Jax steers us in such an inelegant direction but think this is a novel, workable way to deal with the limitations. It also nicely allows a full decoupling between logging and computation (we could even remove the with_mlflow flag as maybe we just don't use the wrapper when we don't want logging?) However, this is a point where we should make a critical decision about switching to Tessseract for general usage which arguably solves the problem much more cleanly with primitives but does not yet support vectorised batching.

adept/adroit.py
```diff
@@ -0,0 +1,102 @@
...
+    def __init__(self, id: str):
+        self.byte_array = jnp.frombuffer(id.encode("UTF-8"), dtype=jnp.int8)
+
+    def __str__(self):
+        return bytes(np.array(self.byte_array)).decode("UTF-8")
+
+    @staticmethod
+    def example():
+        return MLRunId("0" * 32)
```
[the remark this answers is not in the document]
johnbcoughlin: - Yes, I will add a comment to the effect that it's for shape dtype.
- I believe we can rely on it being 32 long, although it might not be part of a formal spec:
```
# Where the DB column is defined [1]:
run_id = Column(String(32), nullable=True, default=None)
...
# In the DB abstraction source file where it's generated [2]:
run_id = uuid.uuid4().hex
```
[1] https://github.com/mlflow/mlflow/blob/b10db98924ec2d6dca7bac44e280ad73244e4f8c/mlflow/store/model_registry/dbmodels/models.py#L97C5-L97C61
[2] https://github.com/mlflow/mlflow/blob/b10db98924ec2d6dca7bac44e280ad73244e4f8c/mlflow/store/tracking/sqlalchemy_store.py#L572

adept/mlflow_logging.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @staticmethod
+    def create_mlflow_run_in_callback(dummy_arg):
+        def create_mlflow_run(dummy_arg):
+            with mlflow.start_run() as run:
+                return MLRunId(run.info.run_id)
+
+        return jax.experimental.io_callback(create_mlflow_run, MLRunId.example(), dummy_arg)
+
+    def call_in_callback(self, callable, *args, mlflow_run_id, **kwargs):
```
jpbrodrick89: callable is a built-in Python function. Also `call_in_callback` is probably too general a name as this is specifically designed for logging using a byte array derived from an MLFLow run id.

Does this need to be a method rather than a module-level function or staticmethod if you prefer (`self` is never used).

PR #160, opened 2025-12-17
feat: add setup() method to MlflowLoggingModule

johnbcoughlin:


jpbrodrick89:
This works, I'm cool with this, just wondering if there's any reason you preferred to either a closure (I guess cumbersome) or to removing the notion of an expected logging pattern entirely (i.e. just use the call back in your call where you like i can sketch this up if it unclear).

johnbcoughlin:
As far as the expected logging pattern, there's no reason we can't break that out further down the line (or right away in a follow-on PR). I think _something_ like this class is nice to have as it codifies how to combine the mlflow logging pattern with an equinox module. Or to put it another way, in a sense we're moving some of the code from `ergoExo.setup` and `ergoExo.__call__` to this class's `__call__` function. Assuming we want to land at a similar logging pattern, that code needs to live somewhere.

Then, assuming we want this class, I prefer the `setup` method to a callback because it just feels like less indirection

PR #161, opened 2025-12-17
feat: Pass parent id through mlflow_logging_module

johnbcoughlin:
Making nested MlflowLoggingModules work requires passing around the run_id as bytes rather than a str


jpbrodrick89:
Haven't looked in detail but how necessary do you think this is? The use would be vmapping over parent runs, not a pattern ive used before but actually might be suitable for Aidan's use case. Just want to make sure we're explicit about the usefulness here for retrospective. If there is just one parent run there is no need for this as a static string is fine.

PR #165, opened 2026-01-06
Use monkey-patch solution instead of mlflow fork

johnbcoughlin:


johnbcoughlin:
I don't _think_ this will break lagradept, but I will follow up straight away there.

PR #166, opened 2026-01-09
feat: mlflow_callback decorator

jpbrodrick89:
This offers some extra flexibility and ease of use if we want to use mlflow logging callbacks without MlflowLoggingModule.

### DIFF
--- a/tests/test_base/test_mlflow_logging.py
+++ b/tests/test_base/test_mlflow_logging.py
@@ -206,7 +206,7 @@ def test_nested_mlflow_logging_modules_vmap(capsys):
 
 def test_mlflow_callback_skips_when_run_id_is_none(capsys):
     @mlflow_callback
-    def log_something(value, mlflow_run_id=None):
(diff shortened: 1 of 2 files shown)


tests/test_base/test_mlflow_logging.py
```diff
@@ -199,3 +199,99 @@ def test_nested_mlflow_logging_modules_vmap(capsys):
...
 
     assert set([m.group(1) for m in pre_logging_matches]) == set([m.group(1) for m in post_logging_matches])
+
+
+# --- mlflow_callback decorator tests ---
+
+
+def test_mlflow_callback_skips_when_run_id_is_none(capsys):
+    @mlflow_callback
+    def log_something(value, mlflow_run_id=None):
```
johnbcoughlin: it doesn't matter except as setting a precedent for this idiom, but it doesn't make sense to use None as a default value here--it will always be non-None if the function is called, as this test asserts.

johnbcoughlin:
this is great, I agree that this is a good direction to go in.

PR #168, opened 2026-01-12
feat: Implement speckled driver

johnbcoughlin:
Implements the LASY speckling envelope.

- RPP and CPP (random phase plate) static speckling
- FM SSD (frequency-modulated spectral dispersion dynamic speckling


johnbcoughlin:
Example run here: https://continuum.ergodic.io/experiments/#/experiments/188384/runs/d09b9af0a83d41c2ac1b9a2c808ceaf2/artifacts

The speckle pattern passes a sanity check for me. For this config we have f_num = 10 and lambda_0 = 351nm, so we should have speckle features on the order of 3.5um.

joglekara (changes requested):
looks fine but probably need to remove the npz stuff?

joglekara (approved):
lgtm, awesome work

PR #178, opened 2026-01-28
chore: Don't pin lineax dep to latest commit

johnbcoughlin:
This fixes incompatible lineax/interpax dependencies that were preventing `uv add git+ssh://git@github.com/ergodicio/adept` from working.


jpbrodrick89:
This sets us back a lot, tridiagonal solves will slow down on GPU, can we not just pin to the SHA just before the commit? Or maybe my jpb/tridiagfromjac branch.

PR #180, opened 2026-01-30
chore: remove sh2d

jpbrodrick89:
This is not used and if we ever want to revive it will probably make more sense to just start from vfp1d, the commits will remain so can always restore directly if we change our mind.


joglekara (approved):
💣

PR #181, opened 2026-01-31
refactor!: consolidate all drift diffusion based collision operators

jpbrodrick89:
I still need to go over the tests, they are purely done by Claude but seem pretty robust, though I haven't even read them! 🙈 

Nevertheless, I'm pretty happy with the overall design which leaves open the door to many opportunities for other models, algorithms and heating/cooling physics.

Key numerical changes **these affect vfp1d, _vlasov1d and vlasov1d2v (vlasov2d unaffected for now)**:
* True zero flux boundary conditions implemented which maximises both Maxwellian preservation and conversation of density. Previously, I think it was effectively Dirichlet which acts as a particle sink. Afaiu zero flux is most commonly used and recommended by Chang-Cooper.
* Central differencing is a _little_ janky due to the drift term being defined on cell edges but I think it has a chance of still being second order accurate (haven't put any thought into this whatsoever), we could just interpolate from edges back to centres but central differencing has loads of bad properties I don't really see why we'd keep it (we can address energy density in other ways going forward).
* Lineax used for all tridiagonal solves (performance issues should be adressed by v0.1.0)

Limitations:
* Currently isn't set up to support spatially varying dv but easy extension (spatially varying nu, C, D are all supported)

Choices:
* Models could maybe be moved to their adept modules as there is no reuse of these
* If you want me to move any files or put in a subdirectory or rename of course more than happy to do that


adept/vfp1d/fokker_planck.py
```diff
@@ -1,100 +1,158 @@
...
+        C_edge = C_edge[0]
+        D = D[0] if D.ndim > 0 else D
+
+        nu_scalar = jnp.array(self.nuee_coeff)
+        op = self.scheme.get_operator(C_edge=C_edge, D=D, nu=nu_scalar, dt=dt)
 
-        :param nu: collision frequency
-        :param f_xv: the distribution function at all locations in space (nx, nv)
-        :param dt: the time step
+        return lx.linear_solve(op, f0, solver=lx.AutoLinearSolver(well_posed=True)).value
```
joglekara: any reason to do this over tridiagonal explicitly?

joglekara (dismissed):
lgtm

tests/test_fokker_planck_utils.py
```diff
@@ -0,0 +1,523 @@
...
+        np.testing.assert_array_equal(C_edge, model.v_edge)
+
+    def test_diffusion_is_energy(self, model):
+        """Test that diffusion coefficient D = <v²>."""
+        # Maxwellian: f = exp(-v²/2)
+        f = jnp.exp(-(model.v**2) / 2.0)
+        f = f / (jnp.sum(f) * model.dv)  # Normalize to unit density
+        C_edge, D = model(f)
+        # For unit Maxwellian, <v²> ≈ 1
+        np.testing.assert_allclose(D, 1.0, rtol=0.05)
```
johnbcoughlin: 0.05 is pretty loose

tests/test_numerical_equilibrium.py
```diff
@@ -0,0 +1,389 @@
...
+    C, D = model(f)
+    print(f"Spherical model D = <v⁴>/<v²>/3 = {float(D):.6f}")
+    print(f"Expected T = {T}")
+
+    # For a Maxwellian: <v⁴>/<v²>/3 = (3T²)/(T)/3 = T  (in 1D)
+    # In 3D spherical: <v⁴> = 15T², <v²> = 3T, so <v⁴>/<v²>/3 = 15T²/3T/3 = 5T/3
+    # Wait, this depends on the measure used in the integral...
+
+
+def test_delta_convention():
```
johnbcoughlin: What is this test trying to validate? There are no assertions.

tests/test_numerical_equilibrium.py
```diff
@@ -0,0 +1,389 @@
...
+    model = SphericalLenardBernstein(v=v, dv=dv)
+    scheme = ChangCooper(dv=dv)
+
+    C, D = model(f)
+    print(f"Spherical model D = <v⁴>/<v²>/3 = {float(D):.6f}")
+    print(f"Expected T = {T}")
+
+    # For a Maxwellian: <v⁴>/<v²>/3 = (3T²)/(T)/3 = T  (in 1D)
+    # In 3D spherical: <v⁴> = 15T², <v²> = 3T, so <v⁴>/<v²>/3 = 15T²/3T/3 = 5T/3
+    # Wait, this depends on the measure used in the integral...
```
johnbcoughlin: ??

johnbcoughlin:
Apologies for commenting on a WIP PR, but I wanted to register my thoughts. I'm looking over the changes to the test structure. I have to say I don't like this style of testing at all. Code reuse in test code, and in particular the use of highly abstracted testing helpers and libraries is a false economy. This is 100x more true with claude code in our toolbelt.

Tests should be readable top-to-bottom, and each test function should test one main behavior. When I suggested testing that the collision operators drive the distribution to a Maxwellian, I had in mind something like this:

```
def test_chang_cooper_fp_equilibrates():
    nu = 0.01
    f0 = initialize_bump_on_tail()
    f_end = solve_fp_relaxation(t_end=1000 * nu, nu=nu, method="ChangCooper")
    assert f_end.iscloseto(maxwellian(f0), tolerance=...)
```

There's a lot of helper code implied in these function calls, yes. But the point is that the top-level test is readable and obvious. With the current structure, it took me several minutes to read up and down the call stack. You have to figure out that the "ends up closer to a Maxwellian" behavior is only tested if you ask for the `"rmse"` "extra_check". And this extra check is only passed along with the Maxwellian initial condition! So we're not testing the property that we wanted at all.

Another lesson in my experience is that full "tensor product" exhaustive test matrices only give you O(epsilon) more information/discrimination than if you'd just tested the corresponding list of outer product factors. The cost of enforcing the full tensor product test suite is that it takes longer to run and you can't manually maintain the list of tests, leading to the sort of abstraction I'm criticizing here. In this PR the test suite is sparsified by the `PROBLEMS` list, but the point is that there's minimal value in defining tests programmatically, so why invest in the machinery to do so?

jpbrodrick89:
I think one easy fix is just removing Maxwelian from the test suite and testing that explicitily. For that case the graphs are of less value. Note that the reason we have to do different tests for that is if it starts off at Maxwellian then therre's no way it's going to end up any closer to Maxwellian!

jpbrodrick89:
Again greatly welcome the feedback and appreciate being challenged on this. Yes, if asked whether this is a sensible test suite to serve us for years to come as our collision operator handling becomes part of the fabric, I'd say "no, probably not". Are these tests overkill? Yes, certainly!

However, I need to iterate fast with different models and schemes for vfp1d and very much "put the code through it's paces" and ensure it behaves expected across a very wide range of problems, be able to compare metrics for all problems in one place and view plots/dashboard every time I make a change. Could I write a script? Yes, but then I'd lose the convenience of pytest iterating and showing assertions. Could I just not commit this? Maybe but then I lose a lot of visibility, and won't be able to share the progress and metrics as well.

If you have a better way to achieve what I'm looking for I'm honestly very open to hearing better ways to achieve this same end. If you're going to squint your eyes and accept this idea id also still appreciate feedback on reducing overengineering and improving understandability. 🙏I must say deleting lines of code with Claude is waaaay harder than creating them. 😅

PR #183, opened 2026-02-02
fix: correct supergaussian parameter naming and thermal velocity calc

johnbcoughlin:
Fixes conflation of mass configuration with the original `m` config value, which was the supergaussian order of the distribution function (`m=2` => Maxwellian).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ license-files = ["LICENSE*"]
 requires-python = ">=3.11"
 dependencies = [
     "lineax",
-    "diffrax @ git+https://github.com/patrick-kidger/diffrax.git@dev",
+    "diffrax",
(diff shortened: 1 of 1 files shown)


johnbcoughlin:
`m` was being conflated between the supergaussian order parameter and the species mass. Basically Claude thought, and I didn't catch, that the original `m` parameter was a mass, when in fact it was the supergaussian order. Then we added an actual mass config for multispecies, and passed that in as `m`. This fixes the error that was introduced.

It looks like the original convention was `v_th = sqrt(T)`, so I'll use `vth = sqrt(T/m)`.

jpbrodrick89:
Lovely makes a lot of sense now sorry for being thick! ☺️ Approved on my side

PR #192, opened 2026-02-10
chore: drop flatdict dependency

jpbrodrick89:
flatdict is no longer maintained and requires setuptools which is no longer packaged with pip.


joglekara (approved):
Pesky legacy hack. Thanks

PR #193, opened 2026-02-12
feat(vfp1d): staggered spatial grids + reflective boundary conditions

jpbrodrick89:
This will eliminate odd-even decoupling while retaining second order differencing of advection terms.

We now support refllective boundary conditions, where f0 is symmetric across a boundary and f1 is antisymmetric (and therefore zero valued.

Also added LLM-generated docs with minor human edits.


joglekara (approved):
very nice!

PR #202, opened 2026-02-20
refactor: some spectrax equinox modules

johnbcoughlin:


joglekara (approved):
is this so you can jit through with your method?

johnbcoughlin:
> is this so you can jit through with your method?

Yes, that's the idea eventually.

PR #204, opened 2026-02-20
cleanup: remove redundant and unused info from spectrax-1d configs

johnbcoughlin:
- Removes the unused `initialization` stanza
- Removes the unused `Ti_Te` field
- The Omega_cs array was redundant with the mi_me ratio. Having both means it's possible to misconfigure something. I replaced the Omega_cs array with a single float `Omega_ce_tau`, which is the nondimensional electron cyclotron frequency. This also standardizes on `me = 1.0`.


joglekara (approved):
ty

PR #212, opened 2026-02-24
Vlasov1D: Parallelizing in x

joglekara:


configs/vlasov-1d/wavepacket.yaml
```diff
@@ -1,6 +1,6 @@
 density:
   quasineutrality: true
-  species-background:
+  species-electron-background:
```
johnbcoughlin: ditto these changes, they're unrelated.

PR #214, opened 2026-02-24
refactor: Begin to define data models for Vlasov1D

johnbcoughlin:
- Introduces regression tests for the configuration logging done by the lifecycle methods that `ergoExo` calls.
- Defines a Vlasov1DSimulation class to serve as the container for all required domain models
- Defines a `Grid` class for the x and t grids. V grids to follow.


joglekara (dismissed):
looks good!

PR #221, opened 2026-02-27
refactor: oshun is dead (and impact too) long live oshun

jpbrodrick89:
IMPACT is not used at all and is largely dead code, rather than try maintain let's just nuke it, mainly because it might be distracting for Claude.

Also, oshun.py and vector_field.py used to be identical but have since diverged. I removed the legacy oshun.py and renamed vector_field.py. If there is a specific reason why you want to call it vector_field.py (i.e. easy identification) I'd be happy to restore that.


joglekara (dismissed):
Sounds good. Might be good to leave it as vector field for CLAUDE to pattern match with the vector fields for all the other solvers but up to you

PR #224, opened 2026-03-02
Vlasov1D - Hou-Li filter

joglekara:


adept/_vlasov1d/datamodel.py
```diff
@@ -120,6 +120,12 @@ class SpaceTermModel(BaseModel):
     width: float
 
 
+class HouLiFilterModel(BaseModel):
+    is_on: bool
+    alpha: float = 36.0
+    order: int = 8
```
johnbcoughlin: In the original paper they choose order=36 as a default which looks pretty good and preserves plenty of modes. 8 seems far too aggressive.

Image

PR #229, opened 2026-03-03
feat(vlasov1d): multiple dist saves per species + diag saves via store_f

joglekara:
New save config format — nested save.<species>.<label> in YAML:

```
save:
  electron:
    main:
      t: {nt: 9}
    full:
      t: {nt: 3}
      x: {xmin: 0.0, xmax: 20.94, nx: 32}
      v: {vmin: -6.4, vmax: 6.4, nv: 512}
```
Each (species, label) pair becomes a flat key (electron.main) used by diffrax SubSaveAt and written to `dist-electron.main.nc`.

Implementation notes:

_species_name is set after get_dist_save_func to avoid polluting the key-set match that selects save mode ({t} vs {t,x,v} vs {t,kx,v})
diag-vlasov-dfdt/diag-fp-dfdt now go through store_f (marked _diag: True) — removed ad-hoc inline storage and dead store_diags
Dot separator chosen over / (breaks netCDF4) and - (ambiguous with old format)
Tests: new test_multi_dist_save.py covers single-label, two-label, file/dataset structure, and full-res saves. 75/75 pass.


johnbcoughlin:
LGTM

PR #244, opened 2026-03-14
refactor(vfp1d): add grid

jpbrodrick89:
I tried to mirror what you guys were doing in Vlasov1D and move towards dependency inject, just a quick pass to check that its in keeping with where you're going and I'm not doing anything silly, cheers

### DIFF
--- a/adept/normalization.py
+++ b/adept/normalization.py
@@ -107,12 +107,4 @@ def skin_depth_normalization(n0_str, T0_str):
     v0 = ((2.0 * T0 / UREG.m_e) ** 0.5).to("m/s")
     x0 = (UREG.c / wp0).to("nm")
 
-    return PlasmaNormalization(
-        m0=UREG.m_e,
-        q0=UREG.e,
-        n0=n0,
-        T0=T0,
-        L0=x0,
-        v0=v0,
-        tau=tau,
-    )
+    return PlasmaNormalization(m0=UREG.m_e, q0=UREG.e, n0=n0, T0=T0, L0=x0, v0=v0, tau=tau)


adept/vfp1d/grid.py
```diff
@@ -0,0 +1,135 @@
...
+        Args:
+            cfg_grid: The ``cfg["grid"]`` dict (raw, with unit strings).
+            norm: Plasma normalization (provides x0, tp0, Te for unit conversion).
+        """
+        xmax = (_Q(cfg_grid["xmax"]) / norm.x0).to("").value
+        xmin = (_Q(cfg_grid["xmin"]) / norm.x0).to("").value
+        tmax = (_Q(cfg_grid["tmax"]) / norm.tp0).to("").value
+        dt = (_Q(cfg_grid["dt"]) / norm.tp0).to("").value
+
+        vmax = 8 * np.sqrt((norm.Te / (csts.m_e * csts.c**2.0)).to("")).value
```
johnbcoughlin: use the v_thermal that has already been calculated?

adept/normalization.py
```diff
@@ -88,3 +94,41 @@ def electron_debye_normalization(n0_str, T0_str):
...
+    Unit quantities are:
+        - c/wp0 (collisionless skin depth)
+        - Electron thermal velocity
+        - 1/wp0
+    """
+    from adept.vfp1d.helpers import calc_logLambda
+
+    T0 = UREG.Quantity(cfg_units["reference electron temperature"])
+    ne = UREG.Quantity(cfg_units["reference electron density"]).to("1/cc")
+    n0 = UREG.Quantity("9.0663e21/cm^3")
```
johnbcoughlin: I don't really understand why the normalization object needs to store both ne and n0. It should only need one number density, which is the density in `1/cc` that "1" means in a simulation. If the actual ne in simulation is different than that, then that's a property of the simulation / initial condition, not the normalization.

adept/normalization.py
```diff
@@ -31,6 +31,12 @@ class PlasmaNormalization:
     # Reference time [s]
     tau: UREG.Quantity
 
+    # Electron density (may differ from normalization density n0)
+    ne: UREG.Quantity | None = None
+    # Coulomb logarithms
+    logLambda_ei_val: float | None = None
+    logLambda_ee_val: float | None = None
```
johnbcoughlin: can these be methods like `logLambda_ee` below, rather than fields?

adept/vfp1d/base.py
```diff
@@ -1,212 +1,146 @@
...
-        ni = ne / self.cfg["units"]["Z"]
-        Te = u.Quantity(self.cfg["units"]["reference electron temperature"]).to("eV")
-        Ti = u.Quantity(self.cfg["units"]["reference ion temperature"]).to("eV")
+        norm = self.plasma_norm
         Z = self.cfg["units"]["Z"]
-        # Should we change this to reference electron density? or allow it to be user set?
-        n0 = u.Quantity("9.0663e21/cm^3")
-        ion_species = self.cfg["units"]["Ion"]
+        ne = norm.ne
+        Te = norm.T0.to("eV")
```
johnbcoughlin: Would be more readable to convert to eV and extract the magnitude on the same line, or name this `Te_eV`.

jpbrodrick89: renamed Te_eV

jpbrodrick89:
Got my head around your design a lot better and like it a lot. Realised we're doing "skin depth' normalization (i.e. v0=c) here and renamed appropriately.

PR #245, opened 2026-03-14
refactor(vfp1d): Fokker Planck dependency injection

jpbrodrick89:
Targeting #244 to minimise diff


jpbrodrick89:
Failing test seems a fluctuation due to an overly sensitive regression test in vlasov1d, this PR shouldnt affect vlasov1d so must be noise I'll let you guys sort that

joglekara (dismissed):
lgtm

PR #248, opened 2026-03-16
Increase resolution of _vlasov1d spacetime plots

johnbcoughlin:


joglekara (approved):
cool. I do like being able to see the whole plot without having to click on it. I'm running into that problem with lagradept so just something to consider.

PR #266, opened 2026-04-08
feat(vfp1d): inverse bremsstrahlung and maxwellian heating/cooling operators

jpbrodrick89:
Maxwellian heating units are still a bit arbitrary and I haven't added time dependent pulse shapes but thought I'd share and propose merging what I have so far.


joglekara (approved):
🚀

PR #279, opened 2026-05-26
feat: osiris wrapper

physicistphil:
Todo:

1. Merge outputs into a single file (xarray) instead of uploading h5
2. Build reasonable diagnostic / plots (mimicking what we have for vlasov). Right now it's just a few useful plots


joglekara:
is this ready to go?

physicistphil:
Gave it a read through; should be good to go. Idk if the tests are actually all that useful but I left them in ¯\_(ツ)_/¯

physicistphil:
### `test-pip-install` failure — not caused by this PR

The `test-pip-install` failure is a pre-existing break on `main` (started 2026-06-15, runs #287/#288; green on 2026-06-10). This branch just inherited it.

**Root cause:** unpinned dependency drift between `matplotlib` and `scienceplots`.

- `import adept` → `adept/_lpse2d/helpers.py` does `import scienceplots`
- `scienceplots 2.2.1` (latest) calls `plt.style.core.read_style_directory(...)` at import time
- `matplotlib 3.11.0` removed the `matplotlib.style.core` attribute, so the import raises:

```
AttributeError: module 'matplotlib.style' has no attribute 'core'
```

`pyproject.toml` pins neither package, and `test-pip-install` does a clean, unconstrained install — so it picked up the freshly released `matplotlib 3.11.0`. The other test jobs use a constrained environment, which is why only this check is red. No fixed `scienceplots` release exists yet (2.2.1 is latest and still references the removed attribute).

**Suggested fix:** pin `matplotlib<3.11` in `pyproject.toml` until `scienceplots` ships a compatible release. Probably best as a small standalone PR against `main` rather than folded in here.

joglekara:
Read through this — looks good to merge from my side. The wrapper architecture is clean, the security surface on the subprocess runner is fine (list-form `Popen`, no shell), and the synthesized-fixture tests (deck roundtrip / post / plots / units) give real CI coverage. One thing I'd like fixed before it goes in:

**Hardcoded local paths in the example configs.** `configs/osiris/twostream-1d.yaml` and `twostream-1d-short.yaml` both point `deck:` and `binary:` at `/home/phil/Desktop/pic/...`, so the shipped examples won't run for anyone else. Since this PR already adds in-repo example decks under `tests/test_osiris/decks/`, could you repoint the configs at those (e.g. `deck: tests/test_osiris/decks/two-stream-1d`) and drop the hardcoded `binary:` in favor of the `OSIRIS_BIN` / `OSIRIS_BIN_<dim>D` env fallback the runner already supports? That makes the examples runnable out of the box.

Otherwise LGTM — nice work on the docstrings and the NetCDF regen path. (Separately: the `test-pip-install` red is the pre-existing matplotlib/scienceplots break on `main`, not this PR.)

physicistphil:
Fixed the hardcoded paths. Also updated the `docs/` markdown file (`osiris-adept-usage.md`). Another note: the canned plots were written for 1D sims, so 2D may cause issues (I haven't tried any 2D runs), but I figure we can cross that bridge when we get to it. Otherwise it should be good to go

joglekara (dismissed):
🚀

physicistphil:
Should be good to go. Recent changes are mostly plotting / quality of life items -- core wrapper remains largely unchanged

joglekara (commented):
Went through this — the code itself looks careful (the crash-salvage logic and the reap/mirror discipline in `stream.py` in particular). The gaps I found are all wiring rather than logic.

## The one thing I'd like before merge: wire the tests into CI

`tests/test_osiris/` is 115 tests across 9 files and none of them have ever run on CI — `.github/workflows/cpu-tests.yaml` has no `osiris` paths-filter and no `test-osiris` job.

The nice thing is that they don't need OSIRIS. I ran the suite at `5bb8dd8` with `OSIRIS_BIN`/`OSIRIS_BIN_1D` explicitly unset:

```
107 passed, 8 skipped in 32.95s
```

The 8 skips are the only ones that genuinely can't run headless — 7 in `test_io_and_plots.py` want a pre-existing run dir (`OSIRIS_TWOSTREAM_RUN`) and exactly 1 wants a built `osiris-1D.e`. Everything else is pure Python: fake bash binaries for the subprocess driver, synthetic HDF5 for the converter/post/plots, and the four checked-in decks for the parser.

| file | tests | covers |
| --- | --- | --- |
| `test_stream.py` | 20 | concurrent H5→NetCDF converter, reap/mirror/quarantine |
| `test_deck_roundtrip.py` | 18 | parser vs. the 4 real decks |
| `test_plots_new_views.py` | 18 | plotting |
| `test_post_netcdf.py` | 17 | post-processing / consolidation |
| `test_adaptive_box.py` | 13 | density-gradient box scaling |
| `test_runner.py` | 7/8 | subprocess driver, error classification, binary discovery |
| `test_units.py` | 7 | skin-depth normalization |
| `test_diagnostics_plots.py` | 7 | diagnostics |

33 s for that is a good trade. The deck parser and the stream converter are the two I most want guarded — a silent regression in either corrupts a real run's data rather than failing loudly.

Second reason: adding the filter folds osiris into the `&shared` anchor, so edits to `adept/_base_.py` and `adept/normalization.py` start triggering it. This PR touches both, which is exactly the coupling worth catching.

Should be ~15 lines copying the `pic1d` block.

## Two papercuts (happy for these to be follow-ups)

**`.gitignore` ignores directories nothing writes to.** It adds `osiris_runs/` and `scratch/`, but the default `run_root` is `./checkpoints` everywhere — `runner.py:120`, `base.py:175`, both configs, and the docs. `checkpoints/` isn't ignored, so a local run leaves untracked GB-scale h5/nc sitting in the repo.

**`docs/osiris-adept-usage.md` is outside the docs build.** It's 348 good lines but it sits outside `docs/source/`, so Sphinx never picks it up, and there's no `docs/source/solvers/osiris/config.md`. Every other solver has one and `CLAUDE.md` asks for it on new config options.

## Notes, not asks

- `h5py` is imported at module scope in `adept/osiris/io.py` but isn't in `pyproject.toml` — it arrives transitively via `h5netcdf`. Works today, silent break if that backend ever changes.
- `stage_discard_h5` (default off) unlinks the scratch HDF5 after `w.append()` but before the NetCDF batch flushes, so a crash inside that window loses those slices from both copies. It's documented as lossy-by-design — just flagging that it's not one to enable for a run that's expensive to repeat.
- `write_units()` returns `{}` for the shipped two-stream deck, since that deck sets neither `n0` nor `omega_p0`, so `units.yaml` comes out empty for the example config. Populates fine on the LPI decks (13 keys). Working as documented, just mildly surprising.

## Things I checked that are clean

- Deck parser round-trips all four real decks exactly at dict level, and no unquoted token gets silently promoted to a quoted string (that was my main worry — it would change deck semantics while still passing the round-trip assertion).
- `subprocess.Popen` takes a list, no `shell=True`, env merge is explicit.
- Every deletion path in `stream.py` early-returns when `persist_dir is None`, so the default non-staged path deletes nothing; the mirror does `copyfile` → `os.replace`.
- The `_base_.py` change is a genuine no-op for the other solvers — `ADEPTModule.__init__` only does `self.cfg = cfg`, so `original_cfg` equals what was being dumped before.

physicistphil:
OSIRIS tests wired in, docs updated/rearranged, and the `.gitignore` has been updated. Should be good to go

joglekara (approved):

PR #280, opened 2026-05-27
deps: drop jpu, use pint directly

joglekara:
## Summary
- `jpu` was used only in [adept/normalization.py](adept/normalization.py) for scalar plasma-normalization arithmetic — no `jpu.Quantity` ever wrapped a traced jax array (every `jnp.*` call already operates on `.magnitude` floats).
- It pinned an upper bound on `jax`. Swapping `UREG = jpu.UnitRegistry()` → `UREG = pint.UnitRegistry()` (pint is already a direct dep) removes the cap with no behavioral change.
- Drop `jpu>=0.0.5` from [pyproject.toml](pyproject.toml) and relock.

## Test plan
- [x] Smoke-test both normalization constructors (`electron_debye_normalization`, `laser_normalization`) plus `normalize`, `vth_norm`, `speed_of_light_norm`, `logLambda_ee`, `approximate_ee_collision_frequency` — all return the same values.
- [x] Import-check every solver helper that depends on `adept.normalization` (`_vlasov1d`, `_vlasov2d`, `_lpse2d`, `_spectrax1d`).
- [x] `pytest tests/test_vlasov1d/test_em_dispersion.py` passes.
- [x] `pre-commit` clean on touched files.

### DIFF
--- a/adept/normalization.py
+++ b/adept/normalization.py
@@ -1,9 +1,9 @@
 from dataclasses import dataclass
 
 import jax.numpy as jnp
-import jpu
+import pint
 
-UREG = jpu.UnitRegistry()
+UREG = pint.UnitRegistry()
 
 
 @dataclass
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,6 @@ dependencies = [
     "pydantic",
     "scienceplots",
     "jax",
-    "jpu>=0.0.5",
 ]
 
 [project.urls]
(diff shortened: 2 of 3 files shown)


jpbrodrick89:
Lgtm but might be worth checking with @johnbcoughlin who introduced the dependency 🙂

johnbcoughlin:
LGTM

PR #304, opened 2026-07-09
feat(hermite-poisson-1d): terms.linear_response — LPSE mode (linearized force, stable at any intensity)

joglekara:
## Why (Scan-5 lesson)

`terms.stabilize` (#303) capped the force on the **full** coupling. In the SRS campaign that converted the strong-forcing detonation into a bounded-coefficient exponential at γ ≈ κ_max·cap ≈ 0.5 ωp — the capped, phase-locked force still pumps the AW ladder — and the above-threshold legs overflowed float64 (fields reached 1e260–1e287 before inf). Lowering the cap enough to kill that rate would bite threshold-relevant fields.

## What

`terms.linear_response: true` removes the mechanism **structurally**: the velocity-space force is linearized about the initialized equilibrium (Maxwellian × density profile),

dC_n/dt|F = κ_n·accel·C_eq_{n−1}

— the textbook −(qE/m)·∂f₀/∂v source, feeding only the momentum row. The force-driven ladder cascade (the entire truncated-AW blow-up channel) no longer exists.

- **Linear kinetics are exact**: Landau damping, χ(k,ω), parametric thresholds and growth rates are linear-response quantities.
- **Above threshold: pure exponential growth, no saturation** — enveloped-LPSE semantics. Auto-enables `wave_density_max = 2.0` so the EM leg saturates gracefully once δn ~ 1 (loop gain collapses to secular growth; no overflow on any practical timescale).
- Runs complete at any intensity; above-threshold amplitudes are not quantitative, by design.

## Tests

- A/B vs the full coupling at weak drive: agree to <1% of the driven response
- survival of the E/α = 1.5 drive that detonates the full coupling at any dt and overflows the tanh-capped variant
- EPW dispersion + Landau damping through the linearized path at standard tolerances

Full HP suite: **35 passed**.


joglekara:
Framing note (from discussion): two clarifications worth recording.

**This is a generally useful mode, not just an SRS-campaign patch.** `linear_response` turns the solver into an exact linearized-kinetics tool: parametric growth-rate and threshold measurement, dispersion/Landau verification, gain spectra in arbitrary density profiles — all with the full-wave EM solver and none of the nonlinear-truncation failure modes. That's a capability LPSE-class envelope codes provide only with modeled damping; here the kinetic linear response is exact.

**Clean growth-to-overflow is not itself a defect — LPSE overflows too** if the growth rate is fast enough or the run long enough; that is the expected endpoint of *any* unsaturated linear model. The distinction that motivated this PR is the *rate*: the tanh-capped full coupling (#303) grew at γ ≈ 0.5 ωp — a numerical artifact rate ~500× the physical SRS γ — putting overflow inside every above-threshold run window. With `linear_response` the growth is at the physical rate, so overflow would require runs ~two orders of magnitude longer than any campaign run: the same practical standing as LPSE past its validity.

joglekara:
Update per review discussion: **`terms.stabilize` / `force_cap` are now removed in this PR** (they merged in #303 but never shipped a working use case — the cap converts the detonation into a γ ≈ 0.5 ωp exponential that overflows in-window, and can't be lowered without biting threshold fields). `linear_response` is the one stable-at-any-intensity mode; `wave_density_max` survives as a plain parameter since it's the graceful-saturation half of LPSE mode and a general wave-equation safety net. Net feature surface: one mode instead of two, no dead knobs. Suite: 32 passed.

PR #310, opened 2026-07-20
fix(vlasov1d): correct v0/L0 normalization (drop spurious sqrt(2))

joglekara:
## Summary

`electron_debye_normalization()` in `adept/normalization.py` defined the velocity unit as `v0 = sqrt(2*T0/m_e)` (most-probable-speed convention), but the entire `_vlasov1d` module — the distribution initializer (`helpers.py`), the Lenard-Bernstein/Dougherty collision operators, the Landau-damping and ion-acoustic dispersion tests, and every shipped example config — actually runs in `v0 = sqrt(T0/m_e)` (RMS/std-dev convention, Maxwellian `∝ e^{-v²/2}` at `T̂=1`).

Numeric (dimensionless) config inputs bypass `normalize()` entirely, so this never corrupted solver dynamics. But every **dimensional-string** input (temperatures, box sizes, gradient scale lengths, laser wavelengths) and every **logged physical unit** (`v0`, `x0`, `c_light`, `box_length`, `nuee`) was off by √2 — e.g. a `normalizing_temperature: 2000eV` run was actually a 4000 eV plasma, and the logged Debye length was √2 too large.

- Drop the `2.0 *` in `v0 = ((2.0 * T0 / UREG.m_e) ** 0.5)` → `v0 = ((T0 / UREG.m_e) ** 0.5)`. `x0` (Debye length) derives from `v0` automatically.
- `vth_norm()` is intentionally **untouched** — it's used only by the unrelated `vfp1d` solver, which is self-consistently built on the `sqrt(2T/m)` convention via its own initializer; changing it would silently halve VFP-1D's simulated temperature.
- Regenerated the 3 `tests/test_vlasov1d/test_config_regression/*` fixture sets (`_units.yml`, `_derived_config.yml`, `_array_config.yml` for `resonance`, `multispecies_ion_acoustic`, `fokker_planck_conservation`) — new values match the corrected convention exactly (e.g. `v0: 26524102 m/s → 18755372 m/s`, `x0: 12.14nm → 8.58nm`, `c_light: 11.30 → 15.98`).
- Documented the normalization explicitly in `docs/source/solvers/vlasov1d/overview.md` so it doesn't silently drift again.

## Test plan
- [x] `tests/test_vlasov1d` full suite (74 passed, 1 xpassed, 7 deselected/GPU) — Landau damping, ion-acoustic, EM dispersion, Fokker-Planck conservation, multispecies, etc. all pass **unchanged**, since they already pinned the RMS convention.
- [x] `tests/test_vlasov1d/test_config_regression.py` regenerated via `--force-regen`, then reran clean (12/12 passed).
- [ ] Downstream consumers of `electron_debye_normalization` (kinetic-srs's `ProbedVlasov1D`, which independently re-derives `v0` for grid pre-normalization) need a matching one-line fix — companion PR: ergodicio/kinetic-srs.

### DIFF
--- a/adept/normalization.py
+++ b/adept/normalization.py
@@ -79,7 +79,10 @@ def electron_debye_normalization(n0_str, T0_str):
     Returns the electron thermal normalization for the given density and temperature.
     Unit quantities are:
         - Debye length
-        - Electron thermal velocity
+        - Electron thermal velocity (RMS/std-dev convention: v0 = sqrt(T0/m_e), so a
+          Maxwellian at T_hat=1 is exp(-v_hat^2/2)) -- this is the convention the
+          _vlasov1d module's initializer, collision operators, and dispersion tests
+          actually run in.
         - Langmuir oscillation frequency
     """
     n0 = UREG.Quantity(n0_str)
@@ -88,7 +91,7 @@ def electron_debye_normalization(n0_str, T0_str):
     wp0 = ((n0 * UREG.e**2.0 / (UREG.m_e * UREG.epsilon_0)) ** 0.5).to("rad/s")
     tau = 1 / wp0
 
-    v0 = ((2.0 * T0 / UREG.m_e) ** 0.5).to("m/s")
+    v0 = ((T0 / UREG.m_e) ** 0.5).to("m/s")
     x0 = (v0 / wp0).to("nm")
 
     return PlasmaNormalization(m0=UREG.m_e, q0=UREG.e, n0=n0, T0=T0, L0=x0, v0=v0, tau=tau)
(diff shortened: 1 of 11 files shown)


joglekara:
Closing in favor of @physicistphil's branch (vlasov1d-conventions-fix), which fixes the same core v0/L0 normalization bug (verified byte-identical regenerated regression fixtures) plus the rest of the units-audit findings this PR intentionally left out of scope (F3-F12), and a real bug the audit missed (logLambda_ee sign error, also duplicated in _tf1d). Phil will submit that as the PR to land instead.

PR #315, opened 2026-07-23
fix(vfp1d): SNB heat-flux diagnostic on a global absolute-energy grid

jpbrodrick89:
## Summary

The VFP-1D SNB nonlocal-heat-flux diagnostic (`adept/vfp1d/heat_flux.py:snb_heat_flux`, added in #312) discretized its energy groups on a **per-cell** `beta = v^2/(2T)` grid, so each group's absolute energy `eps = beta*T(x)` — and hence its mean free path `lambda ~ eps^2` — varied from cell to cell.

This is physically inconsistent: the SNB multigroup diffusion transports each group at **fixed kinetic energy** (energy is conserved as electrons stream, which is exactly what lets the groups decouple in energy and couple only in space). Gridding on the local `beta` makes the spatial stencil couple *different absolute energies* between neighbouring cells — a spurious energy advection (the dropped `-beta d(lnT)/dx d/dbeta` Jacobian term of the `eps -> beta` change of variable) — and drops the suprathermal carriers that stream from hot to cold and set the preheat.

**Effect:** peak flux only mildly affected, but cold-side **preheat under-predicted by ~25% (mild gradients) to ~5x (steep gradients)**. A `beta_max` sweep (20->160) and group sweep (32->128) left it unchanged, confirming it is structural, not a truncation.

## Fix

Rewrite to a **global absolute-energy grid** `eps = v^2/2` (uniform in space, spanning `[0, beta_max * max(T)]`) with the local Spitzer-Harm group weight `W_g(x) = integral of beta^4 e^-beta / 24` over `beta = eps/T(x)`. Each group is now one fixed energy band across all cells, so the plain spatial stencil *is* the fixed-energy transport and the suprathermal carriers stay on-grid everywhere. This matches the validated reference SNB implementation.

Defaults changed: `ngroups` 32 -> 300, `beta_max` 30 -> 20 (the global-energy grid needs a few hundred groups because cold cells sample `beta` coarsely; `beta_max ~ 20` keeps the `beta ~ 16` flux-carrying groups on-grid at the hottest cell).

## Verification

Driven with coefficients that map exactly onto the reference (`nuee_coeff=1, logLam_ratio=4/Z, ni=n/Z`) on the reference's Z=7 Gaussian problem, `q_snb/q_sh` reproduces the validated reference:

| regime | q_snb/q_sh | max abs err vs reference (no l_s) |
|---|---|---|
| local (T:1->1.3) | 0.998 | 9e-6 |
| mild (T:1->3) | 0.465 | 1e-5 |
| steep (T:1->5) | 0.073 | 5e-5 |

The only residual vs the full reference is the electric-field / `l_s` stopping term in the effective diffusion mfp (2nd-order; ~3% at the steep peak, ~0 in the local limit) — a separate refinement that is **intentionally still omitted** here and left for follow-up.

## Tests

- `tests/test_vfp1d/test_spherical.py::test_snb_local_limit` — passes
- `tests/test_vfp1d/test_spherical.py::test_spherical_hotspot` (full-sim end-to-end) — passes

Docs (`docs/source/solvers/vfp1d/config.md`) and the example config (`configs/vfp-1d/spherical-heatflow.yaml`) updated for the new grid semantics and defaults.


joglekara (approved):
ty!

PR #332, opened 2026-08-10
chore: remove the AWS Batch / cloud submit path from adept

joglekara:
## Why

The Batch queue is available to exactly one person, so shipping the submit client here made `adept` carry `boto3` (plus `botocore`, `jmespath`, `s3transfer`) as a hard dependency for everyone else's benefit of nothing. This reverts the surface added in #321.

The submit client continues to live in `continuum-infra/sim-runner/submit.py`, which is where the runtime half (`bootstrap.sh`) already was. A separate private repo for the launcher is the likely long-term home, but that's out of scope here.

## What

| File | Change |
|---|---|
| `adept/cloud.py` | deleted (483 lines — bundle / upload / `submit_job`) |
| `tests/test_cloud/` | deleted (253 lines) |
| `docs/source/usage/cloud.md` | deleted, plus its toctree entry in `docs/source/usage.md` |
| `pyproject.toml` | dropped `boto3`; `uv lock` also drops `botocore`, `jmespath`, `s3transfer` |
| `adept/utils.py` | dropped `download_from_s3`, `upload_dir_to_s3`, and the boto3 branch of `download_file` — the local-artifact path is unchanged |
| `adept/_lpse2d/modules/driver.py` | dropped the s3 branch of `load()`; loading a driver from a local `.pkl`/`.eqx` still works, and no shipped config used the s3 path |
| `tests/test_lpse2d/configs/tpd.yaml`, `README.md`, `docs/source/tests.md` | stale s3 / Batch references |

Net: **+27 / −1001**.

## Verification

- Every `adept` submodule imports. The one failure, `_tf1d/train_damping.py`'s `from utils import misc, plotters`, is pre-existing on `main` and untouched here.
- `tests/test_base` and `tests/test_lpse2d/test_dealias.py` pass (37 passed). The full CI set was still running locally when this was opened — CI runs all suites anyway, since `pyproject.toml`, `uv.lock` and `adept/utils.py` are in the `shared` paths filter.
- Smoke-tested `driver.load()` directly: still reads a local `.pkl` and still raises `NotImplementedError` on an unsupported suffix.
- All pre-commit hooks pass; `make -C docs html` builds with no warnings (no dangling toctree reference).


joglekara:
Companion PR: ergodicio/ml-for-lpi#13.

That repo was the only real consumer of the removed surface, and only indirectly:

- Its S3 code (`analysis/s3_fetch.py` and the `analyze_*`/`figures` scripts) is already self-contained boto3 — it never imported `adept.cloud` or `adept.utils.download_from_s3`. From `adept.utils` it uses only `log_params` and `all_reduce_gradients`, both untouched here.
- It did rely on `boto3` reaching it transitively through adept. ml-for-lpi#13 declares it directly.
- Two of its callers set `drivers.E0.file` to an `s3://` URI. `tools/validate_opt_runs.py` now pre-fetches to a local path; `configs/tpd-scan.yaml` dropped the key (that URI was dead anyway — the run's artifact prefix lists 0 keys).

Hence the second commit here: a URI now fails as a URI instead of falling through to `open()` and surfacing as a missing local file.

PR #337, opened 2026-08-12
feat: vlasov1d broadband laser driver + 1D super-Gaussian width normalization (vlasov1d & pic1d)

ManharGupta25:
Introducing the broadband changes and the alpha_1D fix for super-gaussian experiments in 1d solvers, mainly vlasov1d. broadband-driver lines plot for amplitude, log(amplitude) and phase are here as well. The additions for mean_e2 metrics and params will be in the existing kinetic-srs PR (https://github.com/ergodicio/kinetic-srs/pull/14)

Claude's detailed comment:

### 1. Broadband laser driver for vlasov1d (forward pass)
- New `BroadbandConfig` / `BroadbandDriver`: the pump is a comb of `num_colors` discrete lines at `dw_j = linspace(-delta_omega, +delta_omega, N) * w0`, each an `EMDriver` with its own frequency offset and phase.
- **Amplitudes are sum-normalized**: `A_j = a0 * sqrt(I_j / sum_k I_k)`, so total power equals `base_intensity` regardless of N — and an N=1 broadband driver reduces *exactly* to the monochromatic driver.
- `intensities` / `phases` blocks support `init: uniform | random` with a seeded RNG and an optional `range` (defaults: `(0, 2)` for intensity weights, `(0, 2*pi)` for phases, the Follett 2019 convention).
- `EMDriver` gains `phase` / `dw0`; `IntensityWavelengthDriverConfig` and `AKWDriverConfig` gain `dw0` / `phase` with `0.0` defaults, so **existing monochromatic configs are unchanged**.
- Implementation notes: keyword `case` patterns (pydantic v2 defines no `__match_args__`), `num_colors` as `int`, driver params passed as a `model_dump()` dict so the driver stays grad-safe for a later backward pass.

### 2. 1D super-Gaussian width normalization
- `alpha = sqrt(Gamma(1/m) / Gamma(3/m))` in `_initialize_supergaussian_distribution_`: the realized **variance equals `T0/mass` for every m**, so a species labeled T0 is at temperature T0 for any order (temperature inflation F(m), the init-vs-operator convention split, and the Debye-length consequences).
- Strict no-op at m=2 (both formulas give `sqrt(2)`): all Maxwellian configs, fixtures and tests unaffected.
- Makes the initializer consistent with the Krook target (variance `T0/mass`) and `SuperGaussianDougherty`'s temperature relation.
- Restores `x0 = lambda_D` at all m, so `k*lambda_D` read from code coordinates is exact across an m-scan.
- The `_pic1d` particle loader (both the inverse-CDF and rejection-sampling sites) gets the same one-line fix — no-op at m=2, so Maxwellian PIC runs are unchanged.
- `docs config.md` m-row updated; verified `<v^2>/T0 = 1.000000` for m in {2, 2.5, 3, 4, 5} in both the vlasov1d initializer and the pic1d pdf form.

### 3. Driver line-spectrum diagnostic in post_process
- `plots/drivers/{ex,ey}-lines.png`: per-line intensity (linear + log10) and phase vs `dw_j/w0`, read from the **live `EMDriver` objects** rather than re-derived from config — so optimized line sets from a future backward pass plot with no extra wiring. Skips single-line drivers; wrapped in a guard so a plotting failure can never cost a completed solve its output.

### 4. Misc
- Merged all the updates made to main branch before this. No conflict should be there.
- Comment cleanup in `solvers/pushers/field.py`.

## Validation
- **N=1 broadband ≡ monochromatic, bit-identical**: every probe metric agrees to the last digit (e.g. `probe_pump_fluence_at_left = 75.85687287627245` on both paths) — driver units, F0 scale and envelope handling confirmed clean.
- **Power bookkeeping**: `sum_j I_j = base_intensity` by construction; CPU seed test: per-seed fluence within 0.88% of the monochromatic control, 3-seed mean within 0.21%.
- **Even-comb periodicity** matches prediction: pulse-train recurrence measured 0.4080 ps vs 0.4098 ps predicted (autocorrelation 0.90).
- **Units sentinel post-merge** (on the deployed install): `x0(4 keV) = 4.942252 nm`, `x0(3 keV) = 4.280116 nm`, `v0 = sqrt(T0/m_e)` — #311 conventions intact; monochromatic configs load with `dw0 = 0.0`, `phase = 0.0`. The alpha change composes cleanly with the #311 scheme: #311 removed the global `sqrt(2)` from the unit definition, this removes the m-dependent F(m) from the initializer — together `T_phys = T_label` for all m.
- **alpha_1D**: quadrature on the initializer's exact form gives `<v^2>/T0 = 1.000000` for m in {2, 2.5, 3, 4, 5} with correct T0 scaling; `test_super_gaussian_fp.py` already builds its references with the 1D convention, so the operator tests double as corroboration.
- **In production**: this branch powers the broadband SRS threshold campaign (multi-seed ensembles at L_n = 208/416 µm, 0.25/0.5% bandwidth, run on Perlmutter + Colab).

## Not in this PR
- Backward pass / optimization through `BroadbandDriver` (planned follow-up; the params-as-dict plumbing anticipates it).


joglekara (changes requested):
Nice work — the core of this is sound and the physics checks out. I checked the branch out, ran the suites, and exercised the broadband path directly. Two things need fixing before merge: one is an unrelated regression, one invalidates the headline validation claim. Everything else is cleanup or docs.

## Blocking

### 1. `FokkerPlanckConfig.m` was deleted — collateral from a merge

`adept/_vlasov1d/datamodel.py` drops `m: float = Field(default=2.0, ge=1.0)`, which `c2e79e4` (#313, the super-Gaussian FP operator) added. It isn't mentioned in the PR description and disappeared in one of the three `Merge branch 'main'` commits, so I'm reading it as accidental.

It's silent rather than fatal — `fokker_planck.py:329` reads the raw `cfg` dict, not the validated model, so `type: super_gaussian` runs still pick up `m`. But the `ge=1.0` bound is gone (an `m: 0.5` deck now validates), and the schema no longer documents the option. Should be a pure restore.

### 2. `num_colors: 1` puts the line at `-delta_omega * w0`, not at `w0`

`jnp.linspace(-Δ, Δ, 1)` returns `[-Δ]`, not `[0]`. Measured on this branch:

```
N=1: dw0/w0 = [-0.0025]
N=2: dw0/w0 = [-0.0025, 0.0025]
N=3: dw0/w0 = [-0.0025, 0.0, 0.0025]
```

So the N=1 reduction test isn't testing what the description claims. The amplitude half is genuinely bit-identical — I reproduced `a0 = 0.06783401532993676` on both paths with x64 on, as `_base_.py` enforces — but the single line is red-shifted 0.25%, so this isn't the monochromatic driver.

That matters because the N=1 equivalence is the main evidence that the driver units and `F0` scaling are right. Either special-case `N == 1` to `dw = 0`, or re-run the comparison at `delta_omega: 0` and restate the claim.

## Should fix before merge

### 3. No config documentation

CLAUDE.md requires `docs/source/solvers/vlasov1d/config.md` updates for new config options. `num_colors`, `delta_omega`, `intensities.{init,seed,range,base_intensity}` and `phases.{init,seed,range,base_phase}` are all undocumented — the only doc change here is the `m` row.

### 4. `delta_omega` is a half-width but reads like a bandwidth

`linspace(-Δ, +Δ, N)` means the full width is `2Δ`, and `plot_driver_spectra` labels `2Δ` as "full width Δω/ω0". Someone writing `delta_omega: 0.0025` for a "0.25% bandwidth" run gets 0.5%. The kinetic-srs notes already carry a translation table for exactly this (`0.25% -> 0.00125`), which is a good sign it's a real trip hazard. Worth stating explicitly in the docs, or renaming to `half_width` / switching the config to full width.

### 5. `intensities: dict` and `phases: dict` are unvalidated

Everything else in `datamodel.py` is a typed pydantic model. Here `init: random` without a `seed` raises a bare `KeyError: 'seed'` from inside the constructor, and `base_intensity` is required but undeclared. Two small `BaseModel`s would turn both into proper validation errors.

### 6. `scale_intensities` / `get_partition_spec` / `__call__` are unreachable, and two of them raise

I gather these are intended for the optimization follow-up, so this is about wiring rather than deletion. As written they can't be called from a built simulation: `BroadbandDriver` is a local inside `EMDriver.from_config`, only `broadband_driver.driver_list` escapes, and `EMDriverSet` stores `list[EMDriver]`. Nothing retains the instance — I checked kinetic-srs too (including `mg/vlasov-broadband-and-sup-gauss`); no Python there references these, and the campaign decks set only `init`/`seed`/`base_intensity`.

All three also fail if you do reach them. Calling them directly on a `BroadbandDriver`:

- `__call__` -> `TypeError: unsupported operand type(s) for |: 'dict' and 'SpaceTimeEnvelopeFunction'` (`self.envelope` is a module, not a dict)
- `get_partition_spec` -> `KeyError: 'learned'`
- `scale_intensities` -> `KeyError: 'activation'`

`__call__` also re-applies `jnp.tanh(phases) * jnp.pi` to phases already drawn on (0, 2π), which contradicts what's baked into `driver_list`.

So whichever way you want to go: either keep a reference to the `BroadbandDriver` on `EMDriverSet` and fix the three methods now, or leave the methods with an explicit `raise NotImplementedError` until the backward-pass PR wires them. Right now they read as working API that silently isn't.

### 7. `self.intensities` holds amplitudes, not intensities

It's `a0 * sqrt(w_j / Σw)`, and `__call__` publishes it under the key `"intensities"`. Worth renaming to `amplitudes` (keeping the raw weights separate if the optimizer wants them) before the optimization PR builds on the name.

### 8. No test and no example config

`tests/test_vlasov1d/` gets only the `EMDriver` signature fix. Three cheap ones that would have caught #2:

- uniform N-line comb -> `Σ a_j² == a_mono²` (this is exact today, I verified)
- seeded reproducibility across two builds (also holds today)
- N=1 comb sits at `dw0 == 0`

An example deck under `configs/vlasov-1d/` would help too, since the kinetic-srs configs are gitignored and the YAML shape currently exists only in that repo's NOTES.md.

### Minor

- `# %%` cell markers left at `datamodel.py:124` and `:163`
- `gamma_5_over_m` in `_vlasov1d/helpers.py` is now unused (pic1d switched imports; vfp1d has its own)
- `match cfg.params` has no `case _:` — a future params type would silently return `None` rather than raise
- `plot_driver_spectra` hardcodes driver key `"0"` when reading `base_intensity`; a driver keyed `'1'` just loses the absolute axis scale (guarded, so cosmetic)

## Physics — this part holds up

**The alpha change is correct and I'd take it on its own.** With `α = sqrt(Γ(1/m)/Γ(3/m))` the variance is `v_th² · (Γ(1/m)/Γ(3/m)) · (Γ(3/m)/Γ(1/m)) = T0/mass` for every `m`. And `SuperGaussianDougherty`'s `D = β^(-2/m) Γ(3/m)/Γ(1/m)` equals exactly that variance — so the initializer is now the operator's fixed point at the labeled temperature, which it wasn't before. `vfp1d` correctly keeps its own 3D-isotropic `sqrt(3Γ(3/m)/(2Γ(5/m)))`, so `configs/vfp-1d/epp-short.yaml` (`m: 3.4`) is untouched. Config-regression and super-Gaussian FP suites pass, and the docs row is updated.

Three notes on it:

1. **It silently changes results for every `m != 2` vlasov1d/pic1d run.** No in-tree vlasov1d config uses `m != 2`, so nothing breaks here — but saved runs and in-flight m-scans are now on a different convention with no version marker. Same class of silent drift as #311. Worth a line in the PR description and wherever the m-scan results are recorded.

2. **All comb lines share `k0`.** The source is `-factor * ω_j² * a_j * sin(k0*x - ω_j*t + φ_j)`, so an off-center line isn't launched on its own dispersion branch. That's fine as used — the campaign decks drive through a `width: 1.0um` point source, where the phase error across the antenna is negligible. It stops being fine for a source extended across the box, where `δk*L ~ 10 rad` at 0.25% over 200 µm. Worth a docstring sentence noting the comb assumes a localized antenna.

3. **"Same power" is time-averaged, not peak.** `Σ a_j² = a0²` exactly (verified), but a phase-locked comb reaches `a0*√N` at recurrence — consistent with the 0.408 ps recurrence you measured, and correct physics for a coherent sum. Just worth confirming the threshold campaign's "same intensity" control means the same thing on both sides of the comparison.

---

Blocking items are only #1 and #2, both small. Happy to look again as soon as those land.

ManharGupta25:
Broadband driver injestion was changed from breaking it into separate EMDrivers in simulation.py to having the injestion being directly vectorized and array-based within field.py such that when complete optimization loop with loss calculation and training wrappers is added, gradient will be traceable back to the amplitudes and phases in transverse wave sources (or any other parameter of those sources which will require some specific plumbing for the given parameter). More importantly, the BroadbandDriver object now persists throughout the simulation as it is directly made part of EMDriverSet, and its the entire EMDriverSet that is stored in 'args'. Along with that, other concerns mentioned in the above PR comment were addressed as well, apart from `#6`, which will be addressed in a larger optimization-related PR.

Claude's detailed comment:

---
## Points addressed — `4c45d6d` + `ea1a394`

Everything is addressed except item `#6` (deferred, see below). The driver was
also restructured beyond the review asks: broadband no longer expands into N
`EMDriver`s — it is now a **single `BroadbandDriver` eqx.Module with `(N,)`
array leaves** (`amplitudes`, `delta_omega`, `phases`, raw `intensity_weights`),
retained on `EMDriverSet.ey`. The pusher reads drivers from `args["drivers"]`
(no silent fallback), computes point-source scales in-trace, and evaluates all
lines vectorized — so the line parameters are differentiable today
(`test_gradients_flow_through_args_route`: finite nonzero grads w.r.t.
amplitudes/phases, both source types). `ea1a394` fixes a latent bug in
`BaseVlasov1D.__call__`'s trainable-module loop (`state` unbound, result unused).

**Blocking:**
- `#1` — `FokkerPlanckConfig.m` restored with `ge=1.0`.
- `#2` — `num_colors: 1` special-cased to `delta_omega = [0]`. Claim restated on
  solid ground: `test_single_line_is_monochromatic` (evaluated source matches
  the mono driver, rtol 1e-12, both source types) and a full-solve control in
  kinetic-srs that reproduces the monochromatic run **bitwise** in every probe
  metric (on a machine verified bitwise run-to-run deterministic).

**Should-fix:**
- `#3` — config.md gains a broadband section: YAML example + full key table +
  how the lines are built.
- `#4` — kept the name, documented hard: half-width stated in docs + both
  docstrings with the "0.25% → 0.00125" example; spacing pinned by test.
- `#5` — `intensities`/`phases` are typed pydantic models; random-without-seed
  and missing `base_intensity` are `ValidationError`s (tested).
- `#7` — renamed to `amplitudes`; raw weights kept as `intensity_weights`.
- `#8` — `test_broadband_driver.py` (8 cases: your three suggested tests +
  half-width, validation, gradient regression) and
  `configs/vlasov-1d/srs-broadband.yaml`.

**Item `#6` — deferred to a dedicated backward-pass PR**, which will add
`intensities.{learned, activation}` / `phases.learned`, point
`get_partition_spec` at the raw leaves, fix `__call__` (incl. dropping the
double `tanh` on phases), and add `init_modules()` so the driver enters the
now-fixed module loop. This PR already provides the retained instance, the
differentiable args route, and the loop fix.

**Minors:** `# %%` markers removed; `case _:` guard added; unused
`gamma_5_over_m` removed; `plot_driver_spectra` no longer hardcodes driver key
`"0"` and reads the array leaves.

**Physics notes:** (1) agreed on the silent `m != 2` change — recorded in the
m-scan campaign notes (old alpha ran m=3/4 at ×1.240/×1.371 the labeled T_e),
isothermal rerun prepared, line going into the PR description; (2) shared-`k0`
/ localized-antenna caveat now in docstrings + docs; (3) time-averaged-vs-peak
documented and *measured*: N=8 comb fluence equals mono within 0.9%/seed
(mean +0.2%) while peak intensity is 2.7–3.9× — and the redesign reproduces the
pre-redesign validation matrix to a few ULP.

PR #342, opened 2026-08-19
Add arbitrary-harmonic 2D3P VFP solver

joglekara:
## Summary

- add a `vfp-2d` ADEPT module for 2D3P electron Vlasov–Maxwell–Fokker–Planck evolution with arbitrary packed complex spherical harmonics
- implement the Tzoufras streaming, electric, and magnetic angular couplings, full 2D Maxwell curls, spectral Poisson initialization, and per-ℓ implicit anisotropic collisions
- keep the JAX/Diffrax state real-valued while restoring complex harmonics inside the operator, with compact array storage suitable for `jit`, `vmap`, autodiff, and sharding
- add spatial inverse-bremsstrahlung heating, Joglekar/KALOS moments, and an experimental long-timescale kinetic-Ohm mode with the 2.5D hidden density-gradient source
- add Landau-damping and reduced Joglekar 2014 configurations, solver documentation, and equation-level/end-to-end tests

## Supporting fix

The existing VFP1D `FLMCollisions` path assumed the ℓ=1 harmonic and real input. This change exposes a per-harmonic solve and handles complex modes by solving their real and imaginary components independently, allowing the same radial collision operator to be reused for every `(ℓ,m)`.

## Impact

This establishes the 2D3P arbitrary-`f_lm` foundation needed for magnetized nonlocal transport, Biermann generation, Nernst advection, anisotropic-pressure effects, and future electron-kinetic/ion-fluid coupling.

The `kinetic-ohm` integrator is intentionally documented as experimental. It removes light/plasma-wave timestep restrictions through an inertia-free moment closure and current projection; a fully implicit kinetic-current response and moving-ion fluid coupling remain follow-on work.

## Validation

```text
.venv/bin/pytest -q tests/test_vfp1d/test_fp_models.py tests/test_vfp2d
............................
```

The reduced `configs/vfp-2d/joglekar-2014-prl.yaml` deck also completed end to end and produced finite fields plus all five generalized-Ohm contributions.


jpbrodrick89:
Just a note that log mean flux only conserves energy exactly near maxwellian, chang cooper will probably be slightly better when far from maxwellian. Energy is only conserved for small timesteps much less than the collision time.

PR #368, opened 2026-09-07
feat(lpse2d): 2-D hybrid particle evolution, exact-rotation light coupling, TPD pump depletion

physicistphil:
Two commits on `lpse2d/srs` since #320.

## 1. 2-D hybrid particle evolution (e7b0903)

Lifts the `ny == 1` restriction on `terms.hpe`. The Follett-2017 test-particle Landau-damping feedback runs in a genuinely 2-D box; quasi-1D is the `n_angles == 1` special case of the same code path.

- The resonance surface ω = k·v is a line in (vx, vy); its integral equals the 1-D formula applied to the projection of f onto each mode's direction. Particles are binned along `n_angles` directions in [0, π) and each mode interpolates between its two bracketing directions.
- Gather indices are static and all angles are binned in one fused scatter, so the per-step cost is flat in `n_angles`; no particle-count increase over 1-D is needed.
- Tail cut on speed in 2-D (projection matches the 1-D truncated Maxwellian above the cutoff); thermalizing x wall gets a cosine-weighted emission direction and tabulated speed law; `y_thermal_frac` thermalizes a fraction of transverse crossings.
- New `terms.hpe.hist_smooth` (binomial filter on the histogram before the slope; default 2 in 2-D, 0 in quasi-1D). Motivation: the γ ≥ 0 clamp pins shot-noise-steep modes to exactly zero, and a zero-damped mode grows relative to the band. Fraction of band modes clamped at hist_smooth 0/1/2/4: 0.061/0.035/0.019/0.004 at 20k particles.
- Latent 1-D fixes: `fhot`/`hpe_mean_energy` used a per-component u instead of |u|; HPE smoke tests now honour the documented NaN sentinel of `hpe_gamma_ratio_kpeak` and take a median over the tail of the run.
- Tests: `tests/test_lpse2d/test_hpe_2d.py` (11 tests: isotropic tail matches the analytic Landau rate with no angular bias; the ky = 0 column of a 2-D box reproduces quasi-1D) plus the updated `test_hpe.py`. Design record in `docs/dev/lpse2d-hpe-plan.md` §7; config `configs/envelope-2d/srs-hpe-2d.yaml`.

## 2. Exact-rotation light coupling, TPD pump depletion, light filter (abe1b9e)

Three opt-in `terms.light` options for the coupled (pump-depletion) light solver, all default off:

- `coupling: rotation`. The MATLAB staggered real/imaginary light update is a leapfrog only for operators of the form i × (real operator). The E0↔E1 coupling through the complex, spatially rotating ∇²φ is not one, so on that term the scheme is forward Euler on a rotation: per-sub-step amplification 1 + sin²(arg ∇²φ)(Ω dt_l)²/2 with Ω = e|∇²φ|/(4mₑ√(ω₀ω₁)), and with propagation in the loop a growth of about 0.04 Ω per unit time independent of dt_l. With ∇²φ frozen over the sub-steps the coupling matrix satisfies M² = −Ω²I, so exp(τM) = cos(Ωτ)I + sin(Ωτ)/Ω·M is applied exactly, Strang-split around the existing propagation update, and conserves ω₁|E0|² + ω₀|E1|² pointwise.
- `tpd_depletion: true`. Adds i e/(4ωₚ₀mₑ)·e^{i(ω₀−2ωₚ₀)t}·ρE_y to dE0_y/dt, the energy-conserving partner of the TPD potential source (the MATLAB original depletes the pump through SRS only, so the EPW energy could exceed the absorbed laser energy).
- `filter: <fraction of π/dx>`. Optional isotropic low-pass on E0, E1 once per EPW step (diagnostic).

`RamanLight.rhs` gains a `couple` flag so the E1 RHS can drop the exchange term when the rotation handles it. Tests in `tests/test_lpse2d/test_light_coupling.py`: explicit-scheme growth for a complex Laplacian, rotation action conservation to 1e-12, small-angle agreement with the explicit scheme, dW_pump = −dW_epw for the TPD pair.

Evidence from controlled runs (srs-campaign `sims/lpse/srs-2d-testbed/NOTES.md`, "Blowup investigation"; MLflow `lpse_2d_testbed` / `lpse_2d_testbed_blowup`): the first 2-D SRS+TPD testbed runs went non-finite at 4–5 ps with the explicit scheme (halving dt_l only moved the NaN from 4.50 to 4.85 ps); `coupling: rotation` runs stay finite to 20 ps; with TPD on but no depletion the EPW held 4× the absorbed laser energy at 4 ps, and `tpd_depletion` holds that ratio at ≤ 0.53.

## Behaviour of existing configs

Unchanged: quasi-1D HPE runs use `hist_smooth: 0` and the `n_angles == 1` path; the three light options default off.

## Test status

- `tests/test_lpse2d/test_light_coupling.py`: 4 passed (CPU, laptop).
- Full `tests/test_lpse2d` suite (`pytest tests/test_lpse2d -x`, CPU jax 0.9.0.1, laptop): 60 passed, 1 skipped, 4 deselected in 5 min 35 s.

## Not in this PR

- The 2-D testbed still shows an unexplained late light-budget effect (reflected flux exceeding the incident from ~4.6 ps, present with TPD off and unchanged by any of the options above). Campaign-side investigation.
- Whether `coupling: rotation` and `tpd_depletion` should become defaults. The rotation scheme still has narrow resonant windows above Ω·dt_l ≈ 0.009 in the periodic model problem; a unitary (spectral or Crank–Nicolson) light propagator would remove them.


physicistphil:
Closing as superseded. Its three commits are on `main` by content already — the 2-D HPE (`e7b0903`) via #363 (`b7787b8`, `c48e2ed`, `789b1c7`) and the exact-rotation light coupling / TPD depletion / light filter (`abe1b9e`) via #369 (`1d229a5`) — and the one piece that was not, the `terms.light.one_way` diagnostic (`9055c36`), is cherry-picked onto the plan-2 branch as `bcee98d` (#378). The branch stays; nothing here needs the conflict resolved.

